PR #1, opened 2019-08-02
bootstrap repository

lieut-data:
This setups the repository, API, and an initial `plugins.json` based on the current set of Mattermost plugins (modulo mattermost-plugin-memes, since it hasn't been released as a monolithic plugin bundle yet).

It does not yet support filtering or the `ServerVersion` argument, but does support pagination.

I credit the previous work done on private cloud to help bootstrap this effort.


internal/model/general.go
```diff
@@ -0,0 +1,6 @@
+package model
```
hanzei: I'm not a fan of `model` packages. Even though this pattern is used in the server, it's very "non go". I would much rather organize code by functionality.

What do you think about moving `plugin.go` into the `api` package (for now)?

internal/model/id.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/pborman/uuid"
+)
+
+var encoding = base32.NewEncoding("ybndrfg8ejkmcpqxot1uwisza345h769")
+
+// NewID is a globally unique identifier.  It is a [A-Z0-9] string 26
+// characters long.  It is a UUID version 4 Guid that is zbased32 encoded
+// with the padding stripped off.
+func NewID() string {
```
hanzei: Why not use `model.NewId()` for this? https://github.com/mattermost/mattermost-server/blob/master/model/utils.go#L140

internal/model/plugin.go
```diff
@@ -0,0 +1,49 @@
...
+
+	mattermostModel "github.com/mattermost/mattermost-server/model"
+)
+
+// Plugin represents a Mattermost plugin in the marketplace.
+type Plugin struct {
+	HomepageURL       string
+	DownloadURL       string
+	DownloadSignature []byte
+	Manifest          *mattermostModel.Manifest
```
hanzei: I'm probably to late with my comment, but why is the whole manifest included in the data struct? I'm concern is that we now have two sources of truth: The data struct and the file in the repo. My proposal would be to have a link to the manifest instead.

cmd/marketplace/server.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/spf13/cobra"
+)
+
+var instanceID string
+
+func init() {
+	instanceID = model.NewID()
+
+	serverCmd.PersistentFlags().String("database", "plugins.json", "The read-only JSON file backing the server.")
+	serverCmd.PersistentFlags().String("listen", ":8085", "The interface and port on which to listen.")
```
marianunez: Just wondering if it's worth having these settings in a config file from the start?

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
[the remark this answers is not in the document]
lieut-data: Actually, I don't think so -- it's in the context of:
>    APPENDIX: How to apply the Apache License to your work.

and instructs on how to apply the license to individual files and such. FWIW, we have exactly the same block on https://github.com/mattermost/mattermost-server/blob/master/LICENSE.txt.

cmd/marketplace/server.go
```diff
@@ -0,0 +1,97 @@
...
+	"os"
+	"os/signal"
+	"time"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-marketplace/internal/api"
+	"github.com/mattermost/mattermost-marketplace/internal/model"
+	"github.com/mattermost/mattermost-marketplace/internal/store"
+	"github.com/pkg/errors"
+	logrus "github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
lieut-data: Oops, unintentional!

internal/api/client.go
```diff
@@ -0,0 +1,65 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/mattermost/mattermost-marketplace/internal/model"
+	"github.com/pkg/errors"
+)
+
+// Client is the programmatic interface to the provisioning server API.
+type Client struct {
```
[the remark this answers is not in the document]
lieut-data: At the moment, only testing: but I think it's generally a good practice to expose a programmatic interface to the API, especially in an open source project meant for reuse.

internal/api/context.go
```diff
@@ -0,0 +1,28 @@
+package api
+
+import (
+	"github.com/mattermost/mattermost-marketplace/internal/model"
+	"github.com/sirupsen/logrus"
+)
+
+// Store describes the interface to the backing store.
+type Store interface {
```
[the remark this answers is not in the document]
lieut-data: The codebase is too small for me to make a convincing argument yet, but I generally advocate for the strategy described at https://blog.chewxy.com/2018/03/18/golang-interfaces/ (and elsewhere): namely, define the interface at point of use, and eschew the "Java-style" interface usage.

If the store grows to include methods not required by the API, the benefit is that this interface stays small. The downside is that if we end up mocking the store in lots of places, we potentially duplicate the interfaces. I'm inclined to decide if the current approach is suboptimal later.

LICENSE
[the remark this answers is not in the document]
hanzei: Well, it just caught my eye. If this is the same in the server, I'm fine with leaving as it is.

internal/api/client.go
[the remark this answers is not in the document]
hanzei: Thanks for the explanation. Maybe it's worth to move this into into it's own package? That would make clear that this is not part of the server code.

hanzei:
`internal` packages can't be imported from other projects. IMO it's good practice to use for code that should be used by others. See https://golang.org/doc/go1.4#internalpackages for more information about this.

internal/model/general.go
lieut-data: Agree that this is probably non-go, but given we use the `model` structs in both `api` and `store`, it seems weird to make `store` depend on `api`.

I might perhaps push for just flattening `internal` altogether, but that just kicks the problem down the road. Do you know of another project with similar fundamentals that organizes itself in a more idiomatic way?

See also https://github.com/mattermost/mattermost-cloud/tree/master/internal to which we could make an argument for Mattermost consistency (though I note I contributed this layout to both).

internal/model/plugin.go
lieut-data: We need the manifest "up front" to enable the marketplace and potentially the server to make informed decisions about a plugin without having to download anything. For example, we have the existing `min_server_version` which would be helpful in driving decisions about what to display. We're also looking at a `required_config` setting.

cmd/marketplace/server.go
lieut-data: I'm inclined to adopt the [12-factor](https://12factor.net/config) school of thought here and prefer the configuration to be defined in the environment vs. a config file. Of course, missing is the ability to actually employ the environment here, but that's basically just a `viper` call away in a subsequent PR. Hoping we can avoid the configuration file altogether, at least in terms of what's checked into the repository.

lieut-data:
@marianunez, see the private cloud repo, but basically `internal` clearly signals to other Go packages that the code inside isn't (yet) meant for public consumption -- it's actually enforced by the Go compiler.

internal/api/client.go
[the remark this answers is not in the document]
lieut-data: Of course, we'll likely want to lift up the client out of `internal` to make it even possible to consume this.

marianunez:
> @marianunez, see the private cloud repo, but basically `internal` clearly signals to other Go packages that the code inside isn't (yet) meant for public consumption -- it's actually enforced by the Go compiler.

Thanks @lieut-data and @hanzei. Still learning my ways around Go 👍

internal/api/client.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/mattermost/mattermost-marketplace/internal/model"
+	"github.com/pkg/errors"
+)
+
+// Client is the programmatic interface to the provisioning server API.
```
ali-farooq0: nit: comment needs updating `provisioning`

internal/model/general.go
hanzei: > it seems weird to make `store` depend on `api`.

Good point :+1: 

> Do you know of another project with similar fundamentals that organizes itself in a more idiomatic way?

I've been searching for idiomatic package layout for a while now, but have found anything great yet.

Maybe move `plugin.go` into it's own package (`plugin`)?

internal/model/plugin.go
hanzei: >  potentially the server to make informed decisions about a plugin without having to download anything

Why is this a requirement?

lieut-data: At a minimum, the marketplace needs to be able to filter down plugins to only show versions applicable to the requesting server in question. It won't be able to do that without the corresponding bit from the manifest. We of course need the id, name, description too. We might want the new `required_config` bit at some point. 

At this point, it seems simpler to just include the whole manifest than conditionally throw things away. Thoughts?

internal/model/general.go
lieut-data: I guess that would just be renaming `model` to `plugin` at this point? I'm hesitant to make this change right now, since the benefit to deviating from `model` isn't quite clear to me yet. I remain open to discussion :)

internal/model/plugin.go
hanzei: My concern was that the JSON DB would get out of sync with the repo. But given that we update the DB after every release of a plugin (right?), this should not be a problem.

I thing this is fine for now. Maybe I'm thinking a bit to much about how this will play out in the future.

PR #2, opened 2019-08-20
support filtering plugins

lieut-data:
Very rudimentary support for filtering plugins via the `filter` query string parameter.

Note that paging already existed. We'll defer worrying about the Mattermost server version for now.

https://mattermost.atlassian.net/browse/MM-17544

### DIFF
--- a/internal/store/plugin.go
+++ b/internal/store/plugin.go
@@ -1,26 +1,60 @@
 package store
 
 import (
+	"strings"
+
 	"github.com/mattermost/mattermost-marketplace/internal/model"
 )
 
+func pluginMatchesFilter(plugin *model.Plugin, filter string) bool {
+	filter = strings.ToLower(filter)
+	if strings.ToLower(plugin.Manifest.Id) == filter {
+		return true
+	}
+
+	if strings.Contains(strings.ToLower(plugin.Manifest.Name), filter) {
+		return true
+	}
+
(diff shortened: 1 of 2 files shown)


internal/store/plugin_test.go
```diff
@@ -83,14 +91,78 @@ func TestPlugins(t *testing.T) {
...
+
+	t.Run("id match, exact", func(t *testing.T) {
+		actualPlugins, err := sqlStore.GetPlugins(&model.PluginFilter{PerPage: model.AllPerPage,
+			Filter: "com.mattermost.demo-plugin",
+		})
+		require.NoError(t, err)
+		require.Equal(t, []*model.Plugin{demoPlugin}, actualPlugins)
+	})
+
+	t.Run("id match, case-insensitivei", func(t *testing.T) {
```
marianunez: Minor typo

internal/store/plugin.go
```diff
@@ -1,26 +1,60 @@
...
+		for _, plugin := range store.plugins {
+			if !pluginMatchesFilter(plugin, filter) {
+				continue
+			}
+
+			plugins = append(plugins, plugin)
+		}
+	}
+
+	if len(plugins) == 0 || pluginFilter.PerPage == 0 {
```
marianunez: Should we move this to the top before filtering? Just in case of the very rare case it requests 0 per page, it does not do an unnecessary pass through all the plugins to filter to later not return any

PR #5, opened 2019-08-22
Developing instructions

ali-farooq0:


README.md
```diff
@@ -15,3 +15,39 @@ This repository houses the open-source components of the Mattermost Integrations
...
+
+1. Install [Go](https://golang.org/doc/install)
+
+### Building
+
+Simply run the following:
+
+```
+# Make sure Go modules is turned on
+$ export GO111MODULE=on
```
hanzei: What do you think about just requiring to clone this outside of `$GOPATH`?

lieut-data: Alternatively, would it make sense to codify this into a `Makefile` as most of our projects seem to end up adopting?

README.md
```diff
@@ -15,3 +15,39 @@ This repository houses the open-source components of the Mattermost Integrations
...
+$ go install ./cmd/marketplace
```
lieut-data: As an aside, I rarely `go install` while /developing/, preferring instead to just `go run ./cmd/marketplace` and get the benefit of one less command while iterating. Would it make sense to distinguish developer from production environments?

README.md
ali-farooq0: Using the `Makefile` totally makes sense, especially since we're using it in other places.
Thinking about keeping this as is till we employ the `Makefile`, unless folks feel we should add it now.

thoughts @hanzei @lieut-data?

README.md
ali-farooq0: Make sense, any benefit in keeping them separate? thinking of just combining the two, since there is no variation in how server is invoked in development vs production environments.

README.md
marianunez: +1 on the Makefile

PR #6, opened 2019-08-28
Expose marketplace client

ali-farooq0:


client/plugins_request.go
```diff
@@ -1,4 +1,4 @@
-package api
+package client
```
hanzei: What do you think about moving the content of this file into `plugins.go`?

client/common.go
```diff
@@ -0,0 +1,37 @@
+package client
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// Client is the programmatic interface to the marketplace server API.
+type Client struct {
```
hanzei: Maybe we should _not_ export this. This was we can make sure, all instances are created through `NewClient`.

lieut-data: That would would trigger a `go lint` warning: https://github.com/golang/lint/issues/210. It would also suppress `Client` from godoc.

client/plugins_test.go
```diff
@@ -1,49 +1,22 @@
...
-
-	return api.NewClient(ts.URL), ts.URL, func() {
-		ts.Close()
-	}
-}
-
 func TestPlugins(t *testing.T) {
 	t.Run("no plugins", func(t *testing.T) {
-		client, _, tearDown := setupApi(t, nil)
+		c, _, tearDown := SetupApi(t, nil)
```
lieut-data: Alternatively, `marketplaceClient`?

client/common_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-marketplace/client"
+	"github.com/mattermost/mattermost-marketplace/internal/api"
+	"github.com/mattermost/mattermost-marketplace/internal/model"
+	"github.com/mattermost/mattermost-marketplace/internal/store"
+	"github.com/mattermost/mattermost-marketplace/internal/testlib"
+	"github.com/stretchr/testify/require"
+)
+
+func SetupApi(t *testing.T, plugins []*model.Plugin) (*client.Client, string, func()) {
```
lieut-data: Why is this exported?

client/common.go
hanzei: Ah, I know there was a reason this isn't a common pattern. As much as I like to reduce the API surface, it's probably not worth the hassle with `go lint`.

But suppressing it from godoc is a plus in my opinion, because it's not important what the fields are.

jasonblais:
cc @ali-farooq0 quick reminder to help review the feedback above :)

ali-farooq0:
Closing this PR as exposing marketplace client isn't immediately necessary. Will turn this into a HW ticket.

lieut-data:
@ali-farooq0, I'd propose we defer turning this into a help-wanted ticket right now, at least until we have a better story around the duplication problem and dependency on the `model` package.

PR #7, opened 2019-08-29
introduce generator cmd

lieut-data:
Add a new generator command for making plugins.json, querying a fixed list of Mattermost repositories, fetching the latest release, downloading assets, and extracting the manifests.


cmd/generator/main.go
```diff
@@ -0,0 +1,219 @@
...
+		if path.Base(hdr.Name) != "plugin.json" {
+			continue
+		}
+		manifest := mattermostModel.ManifestFromJson(bundleReader)
+
+		plugin.Manifest = manifest
+		break
+	}
+
+	if plugin.Manifest == nil {
```
marianunez: This is great that is validating the bundle before going to the marketplace. Will we be adding signature verification as well?

lieut-data: That's a really good idea! No reason to distribute something that fails validation from the get go.

PR #8, opened 2019-09-03
MM-18097: enable running as a Lambda function

lieut-data:
To facilitate easy deployment for the Marketplace MVP, enable the server to run as a Lambda function, compiling the `plugins.json` database directly into the binary.

Fixes: https://mattermost.atlassian.net/browse/MM-18097


notes
```diff
@@ -0,0 +1,6 @@
+Meeting summary:
```
iomodo: Do we need this file?

cmd/lambda/main.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+}
+
+func newStatikStore(logger logrus.FieldLogger) (*store.Store, error) {
+	statikFS, err := fs.New()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to open statik fileystem")
+	}
+
+	database, err := statikFS.Open("/plugins.json")
```
iomodo: nit: I'd pass "/plugins.json" as an argument to the `newStatikStore` function

notes
lieut-data: Oops, nope! That represents some pending action items on my part re: devops. Thanks for spotting :)

PR #9, opened 2019-09-10
Remove circlci badge

hanzei:
Because we are using GH actions ATM


hanzei:
Screenshot from 2019-09-10 10-16-01

CI seams to be broken. I will look into it.

hanzei:
Fix is in #10

PR #10, opened 2019-09-10
Fix CI

hanzei:
- Fix broken git.apache.org repository
- Fix `make generate`


hanzei:
@marianunez @lieut-data would be great if you could review this today, so CI is stable again

Makefile
```diff
@@ -7,7 +7,7 @@ all: check-style build
 .PHONY: generate
 generate:
 	go get github.com/rakyll/statik
-	cp plugins.json data/static/
+	cp plugins.json data/statik/
```
lieut-data: Can you elaborate what's being fixed here? The `data/static` path is the "raw data" and the `statik` package is what we generate. With this change, we'd include not only `plugins.json` but also the `statik.go` that is generated, which doesn't seem right?

hanzei: The problems was that `data/static` didn't exist for the CI and hence the copy fails. But you are correct, that my solutions was to brought. Fix this. Please take a seconds look.

PR #11, opened 2019-09-10
[MM-18470] Add healthcheck api endpoint

hanzei:
This PR adds a `/health` API endpoint. It strictly follows https://tools.ietf.org/id/draft-inadarei-api-health-check-01.html.

I still have to add a test for this, but early feedback is appreciated.

Fixes: https://mattermost.atlassian.net/browse/MM-18470


internal/api/health_check.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/gorilla/mux"
+)
+
+var buildTag = ""
+var buildHash = ""
+
+type healthCheckResponse struct {
+	Status      string
+	Version     string
+	ReleaseID   string
```
hanzei: This is the last git tag. Empty for now

internal/api/health_check.go
```diff
@@ -0,0 +1,38 @@
...
+	Notes       []string
```
hanzei: This contains the sha1 of the last commit. I didn't found a more elegant solution to add this information to the response.

internal/api/health_check.go
```diff
@@ -0,0 +1,38 @@
...
+	Description string
+}
+
+func initHealthCheck(apiRouter *mux.Router, context *Context) {
+	addContext := func(handler contextHandlerFunc) *contextHandler {
```
hanzei: I dislike that I had to re-add this here. Maybe it's worth to refactor into it's own method.

internal/api/plugins.go
```diff
@@ -59,5 +59,6 @@ func handleGetPlugins(c *Context, w http.ResponseWriter, r *http.Request) {
 		plugins = []*model.Plugin{}
 	}
 
+	w.Header().Set("Content-Type", "application/json")
```
hanzei: This fixes the `Content-Type` for the get plugins endpoint.

hanzei:
@lieut-data I've added a test. PTAL

internal/api/health_check.go
ali-farooq0: spoke about it offline, suggested to use `Details` instead of `Notes`

hanzei:
@ali-farooq0 I've made the changes. Please take a second look.

PR #12, opened 2019-09-11
MM-17846: add support for plugin icons

lieut-data:
Currently just embedding icons from https://simpleicons.org/, but we'll likely want to tune this.

https://mattermost.atlassian.net/browse/MM-17846


cmd/generator/main.go
```diff
@@ -87,6 +91,20 @@ var generatorCmd = &cobra.Command{
 			"mattermost-plugin-nps",
 		}
 
+		type iconURL struct {
+			URL    string
+			Inline bool
```
hanzei: Why do you want to support both passing the URL to the server and passing data to the server? Why not stick with one way?

cmd/generator/main.go
```diff
@@ -4,15 +4,19 @@ import (
...
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"path"
 	"strings"
 
 	"github.com/google/go-github/v28/github"
+	"github.com/h2non/filetype"
+	svg "github.com/h2non/go-is-svg"
```
hanzei: You will need to add this to `go.mod`

cmd/generator/main.go
```diff
@@ -87,6 +91,20 @@ var generatorCmd = &cobra.Command{
 			"mattermost-plugin-nps",
 		}
 
+		type iconURL struct {
+			URL    string
+			Inline bool
+		}
+
+		iconURLs := map[string]iconURL{
+			"mattermost-plugin-aws-SNS":        iconURL{"https://unpkg.com/simple-icons@latest/icons/amazonaws.svg", true},
```
marianunez: Should we change this to point to `integrations.mattermost.com`?

lieut-data: This is just for the `generator` script, which pulls down these assets and embeds the base64 data into the `plugins.json`. The server won't ever see or use these URLs.

cmd/generator/main.go
lieut-data: As discussed, sticking with the single way.

PR #15, opened 2019-09-18
Updated plugin icons

lieut-data:
I'm still pending two icons, but this updates icons for various plugins to use their "official" versions.

As part of this change, I've updated the generator tool to:
* subset to the set of prepackaged plugins, subject to [discussion](https://community.mattermost.com/core/pl/had54684gbnsfrrszaroigfmqc)
* add an `--existing` flag to speed up regenerating `plugins.json`
* support ingesting plugin icons from the local path into the `plugins.json` generated file

I'm fully anticipating the generator tool to be heavily refactored into something that will enable us to better manage the database in the long term, so I'm not spending /too/ much time on the architecture.

I'll update this PR with the final two icons once received, but I think we're in a good position to kickstart review.


cmd/generator/main.go
```diff
@@ -253,14 +273,24 @@ func getLatestRelease(ctx context.Context, client *github.Client, repoName strin
...
+	if strings.HasPrefix(icon, "http") {
+		logger.Debugf("fetching icon from url %s", icon)
+
+		resp, err := http.Get(icon)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to download plugin icon at %s", icon)
+		}
+		defer resp.Body.Close()
+
+		return ioutil.ReadAll(resp.Body)
```
marianunez: Not sure if it makes any difference but should this return `nil` for err: 

```suggestion
		return ioutil.ReadAll(resp.Body), nil
```

lieut-data: `ioutil.ReadAll` is [defined](https://golang.org/pkg/io/ioutil/#ReadAll) as the following, so I'm effectively returning any error from reading the body:
```go
func ReadAll(r io.Reader) ([]byte, error)
```

lieut-data:
Good catch, @iomodo -- fixed!

lieut-data:
Thanks! Holding on merging until we have the final two icon assets.

PR #16, opened 2019-10-04
[MM-18800] Add signatures in the plugin.json

iomodo:
This PR
1. Adds signatures field to /plugins API
2. Updates generator to fetch signatures from github and encode it in plugins.json

Fixes: 
https://mattermost.atlassian.net/browse/MM-18800
https://mattermost.atlassian.net/browse/MM-18862

Testing steps:
1. `generator` command should generate correct `plugins.json` file. If Github repos contain signatures they should be in `plugins.json` as well. Signatures should be stored using base64 encoding.
2. Signatures should be returned on /plugins endpoint.


internal/store/plugin_test.go
```diff
@@ -13,10 +13,9 @@ import (
 
 func TestPlugins(t *testing.T) {
 	demoPlugin := &model.Plugin{
-		HomepageURL:       "https://github.com/mattermost/mattermost-plugin-demo",
-		IconData:          "icon-data.svg",
-		DownloadURL:       "https://github.com/mattermost/mattermost-plugin-demo/releases/download/v0.1.0/com.mattermost.demo-plugin-0.1.0.tar.gz",
-		DownloadSignature: []byte("signature"),
+		HomepageURL: "https://github.com/mattermost/mattermost-plugin-demo",
+		IconData:    "icon-data.svg",
+		DownloadURL: "https://github.com/mattermost/mattermost-plugin-demo/releases/download/v0.1.0/com.mattermost.demo-plugin-0.1.0.tar.gz",
```
lieut-data: Should we add some tests in the store for the new `Signatures` field?

internal/model/plugin.go
```diff
@@ -7,13 +7,19 @@ import (
 	mattermostModel "github.com/mattermost/mattermost-server/model"
 )
 
+// PluginSignatures represents Mattermost plugin signatures
+type PluginSignatures struct {
+	Signature     string `json:"signature"`
+	PublicKeyHash string `json:"public_key_hash"`
```
lieut-data: Can we GoDoc this field, in particular, to document the hashing algorithm used?

internal/model/plugin.go
```diff
@@ -7,13 +7,19 @@ import (
...
-	HomepageURL       string                    `json:"homepage_url"`
-	IconData          string                    `json:"icon_data"`
-	DownloadURL       string                    `json:"download_url"`
-	DownloadSignature []byte                    `json:"download_signature"`
-	Manifest          *mattermostModel.Manifest `json:"manifest"`
+	HomepageURL string                    `json:"homepage_url"`
+	IconData    string                    `json:"icon_data"`
+	DownloadURL string                    `json:"download_url"`
+	Manifest    *mattermostModel.Manifest `json:"manifest"`
+	Signatures  []*PluginSignatures       `json:"signatures"`
```
lieut-data: Should we converge on `[]PluginSignatures` in this case?

internal/model/plugin.go
```diff
@@ -7,13 +7,19 @@ import (
 	mattermostModel "github.com/mattermost/mattermost-server/model"
 )
 
+// PluginSignatures represents Mattermost plugin signatures
```
lieut-data: nit: this comment, while technically accurate, seems unhelpful. Perhaps:
```suggestion
// PluginSignatures is a list of public key signatures and the corresponding public key hash for use in verifying a plugin downloaded from the marketplace.
//
// A plugin may have multiple signatures if it is signed by multiple public keys known to be trusted by different versions of the Mattermost server.
```

cmd/generator/main.go
```diff
@@ -246,6 +255,57 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
+		if err != nil {
+			return nil, err
+		}
+		sig, err := getSignatureFromSignatureFileName(asset)
+		if err != nil {
+			return nil, err
+		}
+
+		signature := &model.PluginSignatures{
+			PublicKeyHash: hash,
```
lieut-data: super nit: these fields are out of order relative to the struct.

cmd/generator/main.go
```diff
@@ -246,6 +255,57 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
+	}
+	return signatures, nil
+}
+
+func getHashFromSignatureFileName(asset github.ReleaseAsset) (string, error) {
+	name := asset.GetName()
+	if !strings.HasSuffix(name, ".sig") && !strings.HasSuffix(name, ".asc") {
+		return "", errors.New("signature file has wrong extension")
+	}
+	name = name[:len(name)-4]
```
lieut-data: Perhaps [strings.TrimSuffix](https://golang.org/pkg/strings/#TrimSuffix) instead?

cmd/generator/main.go
```diff
@@ -246,6 +255,57 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
+	lastIndex := strings.LastIndex(name, "-")
+	if lastIndex == -1 {
+		return "", errors.New("can't find public key hash in the signature file name")
+	}
+	return name[lastIndex+1:], nil
+}
+
+func getSignatureFromSignatureFileName(asset github.ReleaseAsset) (string, error) {
```
lieut-data: nit: this doesn't actually get the signature from the /filename/, but from the asset. Perhaps just call this `getSignatureFromAsset`?

internal/model/plugin.go
iomodo: Yep, I was going for consistency here. 
In this concrete case either will work - copying 4 arguments vs dereferencing 2. 
But in general I don't think it's a good idea using `[]struct` everywhere.
1/5 that we should stick to `[]*PluginSignatures`

cmd/generator/main.go
iomodo: I have to check the suffix with `HasSuffix` and after that call the `TrimSuffix` which will once again call the `HasSuffix` and after that return `name[:len(name)-len(suffix)]`. I did not like that . 
Probably better idea just to add a comment that we are trimming the suffix?

cmd/generator/main.go
```diff
@@ -246,6 +255,57 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
+func downloadSignatures(assets []github.ReleaseAsset) ([]*model.PluginSignature, error) {
+	signatures := make([]*model.PluginSignature, 0, len(assets))
+	for _, asset := range assets {
+		hash, err := getPublicKeyHashFromAsset(asset)
+		if err != nil {
+			return nil, errors.Wrap(err, "Can't get public key hash from the asset")
+		}
+		sig, err := getSignatureFromAsset(asset)
+		if err != nil {
+			return nil, errors.Wrap(err, "Can't get signature from the asset")
```
ali-farooq0: same comment as above.

internal/model/plugin.go
ali-farooq0: i wish we had used non-pointers in our other codebase :(, i miss just using `[]struct` :( for simplicity.

ali-farooq0:
@iomodo Should this be merged into `plugin_signing` feature branch? Will the generator fail if it doesn't find signature files or will it just ignore? If it won't fail then we can go to `master` directly.

ali-farooq0:
oh looks like it'll just skip the downloading signatures if it doesn't find any `.sig` or `.asc` files, in which case we're good 👍

internal/model/plugin.go
lieut-data: Agree we should reach for consistency within a repository at a minimum. I had been under the impression we were already using vanilla structs, but looks like I'm wrong, so no changes required here.

As I understand, it's hard to reason about Go's handling of structs passed by value, since it may choose to promote to the heap anyway or it might take optimization steps to avoid even the expense of copying. So if we had to make a decision today, I'd say `[]struct` and trust the Go compiler. 

I'd be open to "fixing" this repository as such and starting a new convention, since I don't think `[]*struct` is a coding style thing that we have to enforce.

cmd/generator/main.go
lieut-data: Personally, `HasSuffix` and `TrimSuffix` are much more clear than a comment and `name[:len(name)-4]`. I think we should optimize for readability over theoretical performance here.

cmd/generator/main.go
```diff
@@ -255,41 +255,41 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
 	name := asset.GetName()
 	if !strings.HasSuffix(name, ".sig") && !strings.HasSuffix(name, ".asc") {
 		return "", errors.New("signature file has wrong extension")
 	}
-	name = name[:len(name)-4]
+	name = name[:len(name)-4] //Trim the suffix
 	lastIndex := strings.LastIndex(name, "-")
 	if lastIndex == -1 {
-		return "", errors.New("can't find public key hash in the signature file name")
+		return "", errors.New("can't find public key hash in the signature file name " + name)
```
lieut-data: nit: use [Errorf](https://godoc.org/github.com/pkg/errors#Errorf)

internal/model/plugin.go
```diff
@@ -7,9 +7,12 @@ import (
...
 )
 
-// PluginSignatures represents Mattermost plugin signatures
-type PluginSignatures struct {
-	Signature     string `json:"signature"`
+// PluginSignature is a public key signature of a plugin and the corresponding public key hash for use in verifying a plugin downloaded from the marketplace.
+type PluginSignature struct {
+	// Signature represents a signature of a plugin saved in base64 encoding.
+	Signature string `json:"signature"`
+	// PublicKeyHash represents first arbitrary number of symbols of the
```
lieut-data: Having trouble parsing this comment: what does `first arbitrary number of symbols` mean?

iomodo: We will check the suffix of the real public key hash. So arbitrary number of symbols might be in `PublicKeyHash` and it will still work. Our own convention will be to use 6 symbols.

lieut-data: Would it make sense to document that `6` here instead? I'm wondering when that constant when change.

iomodo: In the rarest occasion when first 6 symbols of the two public key hashes are the same, we can use first 7 symbols for those keys. It's far-fetched, but still.

On the other hand maybe some of our customers who will have their own marketplace can adopt different number of symbols. They even can adopt variable number of symbols for different keys. They can use 1 symbol of a hash if there is a collision with previous keys add another symbol and so on.

lieut-data: Maybe we can use language like https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection, i.e. `abbreviated public key hashes`.

ali-farooq0:
@iomodo since this is going into `master` does this require sanity testing?

prapti:
@iomodo Can the original description of this PR be updated with a link to the Jira ticket and QA steps to test? That'd be great!

PR #18, opened 2019-10-25
MM-18593: support extra plugin metadata

lieut-data:
Support parsing `IconPath`, `HomepageURL` and `SupportURL` out of a
plugin's manifest. Embed a timestamp as to when the plugin was last
updated. Update to GitHub 0.12.0 with this metadata.

Note that we maintain the old icons for the existing plugins with icons
for when we expand the database to include all versions available.

Fixes: https://mattermost.atlassian.net/browse/MM-18593, https://mattermost.atlassian.net/browse/MM-19602


plugins.json
```diff
@@ -2,13 +2,16 @@
...
+    "download_url": "https://github.com/mattermost/mattermost-plugin-github/releases/download/v0.12.0/github-0.12.0.tar.gz",
     "download_signature": "",
     "manifest": {
       "id": "github",
       "name": "GitHub",
       "description": "GitHub plugin for Mattermost.",
-      "version": "0.11.0",
+      "homepage_url": "https://github.com/mattermost/mattermost-plugin-github",
+      "support_url": "https://github.com/mattermost/mattermost-plugin-github/issues",
+      "icon_path": "assets/icon.svg",
```
lieut-data: Note the embedding of this field, but no change to the icon above -- since v0.12.0 embeds the same file.

cmd/generator/main.go
```diff
@@ -177,9 +182,16 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
+	var updatedAt time.Time
 	for _, releaseAsset := range latestRelease.Assets {
 		if strings.HasSuffix(releaseAsset.GetName(), ".tar.gz") {
 			downloadURL = releaseAsset.GetBrowserDownloadURL()
+			timestampUpdatedAt := releaseAsset.GetUpdatedAt()
+			if timestampUpdatedAt.IsZero() {
+				timestampUpdatedAt = releaseAsset.GetCreatedAt()
+			}
+
+			updatedAt = timestampUpdatedAt.In(timestampUpdatedAt.Location())
```
hanzei: I'm wondering if we not just use `UTC` all the time (haha).

cmd/generator/main.go
```diff
@@ -213,39 +235,73 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
-		bundleReader := tar.NewReader(gzBundleReader)
-		for {
-			hdr, err := bundleReader.Next()
-			if err == io.EOF {
-				break
-			}
-			if err != nil {
-				return nil, errors.Wrapf(err, "failed to read plugin bundle for release %s", releaseName)
-			}
+		bundleData, err := ioutil.ReadAll(gzBundleReader)
```
hanzei: :+1:

cmd/generator/main.go
```diff
@@ -213,39 +235,73 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
 	plugin.DownloadURL = downloadURL
 	plugin.DownloadSignature = []byte{}
+	plugin.UpdatedAt = updatedAt
 
-	if plugin.Manifest == nil {
-		return nil, fmt.Errorf("failed to find plugin manifest for release %s", releaseName)
+	return plugin, nil
+}
+
+func getTarFile(reader *tar.Reader, filepath string) ([]byte, error) {
```
hanzei: It's more like a `getFromTarFile`, right?

cmd/generator/main.go
```diff
@@ -213,39 +235,73 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to read tar file")
+		}
+
+		// Match the filepath, assuming the tar file contains a leading folder matching the
+		// plugin id.
+		if matched, err := path.Match(fmt.Sprintf("*/%s", filepath), hdr.Name); err != nil {
```
hanzei: Nit: I would split this into two line, but this is a personal style choice.

cmd/generator/main.go
```diff
@@ -213,39 +235,73 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
+	}
+
 	// Reset fields, even if we found the existing plugin above.
-	plugin.HomepageURL = repository.GetHTMLURL()
-	plugin.IconData = ""
+	if plugin.Manifest.HomepageURL != "" {
+		plugin.HomepageURL = plugin.Manifest.HomepageURL
+	} else {
+		plugin.HomepageURL = repository.GetHTMLURL()
+	}
```
hanzei: You forgot `SupportURL`. We can use https://godoc.org/github.com/google/go-github/github#Repository.GetIssuesURL for this

cmd/generator/main.go
hanzei: Maybe we want to be explicit here and check for the plugin id? 1/5

cmd/generator/main.go
```diff
@@ -121,20 +123,23 @@ var generatorCmd = &cobra.Command{
...
-				icon, err := getIcon(ctx, iconPath)
-				if err != nil {
-					return errors.Wrapf(err, "failed to fetch icon for repository %s", repositoryName)
-				}
-				if svg.Is(icon) {
-					plugin.IconData = fmt.Sprintf("data:image/svg+xml;base64,%s", base64.StdEncoding.EncodeToString(icon))
-				} else {
-					kind, err := filetype.Image(icon)
+			if len(plugin.IconData) == 0 {
+
```
hanzei: Nit: Remove this line

go.mod
```diff
@@ -10,18 +10,15 @@ require (
 	github.com/gorilla/mux v1.7.3
 	github.com/h2non/filetype v1.0.10
 	github.com/h2non/go-is-svg v0.0.0-20160927212452-35e8c4b0612c
-	github.com/mattermost/mattermost-server v5.12.3+incompatible
+	github.com/jehiah/go-strftime v0.0.0-20171201141054-1d33003b3869 // indirect
+	github.com/mattermost/mattermost-server master
```
hanzei: That is not a correct version. Would you please run `go mod tidy`.

cmd/generator/main.go
lieut-data: Actually, I omitted this on purpose, since it will be captured in the plugin's manifest (when all the plugins are updated). I only "promoted" `HomepageURL` since it's already a field v5.16/v5.17 expects to have at the top-level, outside the manifest.

Perhaps in a `v2` API endpoint, we can strip even this duplication and effectively just deliver manifests.

cmd/generator/main.go
lieut-data: I originally wrote this was the closest thing I could fine to a `Clone` -- just copying the timezone from the incoming timestamp. I'm assuming UTC at the source, but I suppose it doesn't hurt to enforce.

cmd/generator/main.go
lieut-data: In general, I'd agree, but creating another variable name here seems more painful. 1/5.

lieut-data: Interestingly, at the moment I can't know the plugin id until I read `plugins.json`, which requires this generic match anyway.

cmd/generator/main.go
```diff
@@ -196,12 +207,22 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
 		}
 	}
 
-	// If no plugin in existing database, attempt to download and inspect manifest.
-	if plugin == nil {
-		plugin = &model.Plugin{}
+	// If no plugin in existing database or the updated timestamp has changed, attempt to download and inspect manifest.
+	if plugin == nil || updatedAt.IsZero() || plugin.UpdatedAt.Before(updatedAt) {
+		if plugin == nil {
+			logger.Debug("no existing plugin; fetching download url")
```
marianunez: We probably don't need the text "fetching download url" given that is always printed below after this `if`

cmd/generator/main.go
```diff
@@ -213,39 +234,73 @@ func getReleasePlugin(ctx context.Context, client *github.Client, repositoryName
...
+			logger.Debugf("using icon specified in manifest as %s", plugin.Manifest.IconPath)
+			plugin.IconData = fmt.Sprintf("data:image/svg+xml;base64,%s", base64.StdEncoding.EncodeToString(iconData))
 		}
 	} else {
 		logger.Debugf("skipping download since found existing plugin")
 	}
 
+	if plugin.Manifest == nil {
+		return nil, fmt.Errorf("failed to find plugin manifest for release %s", releaseName)
+	}
```
marianunez: Should this check be before we try to get the `IconPath` in:

``if plugin.Manifest.IconPath != "" {``

line 248

lieut-data: We write to `plugin.Manifest` just before that:
https://github.com/mattermost/mattermost-marketplace/blob/9a9787e3bbad25cb45ca6649e3facf94ccdd903b/cmd/generator/main.go#L246

but I suppose this could still return `nil`. I'll add a check.

cmd/generator/main.go
hanzei: Well, that's a blocker. NVM then

cmd/generator/main.go
hanzei: Got it.  :+1: I didn't know `HomepageURL` was a thing before 5.18.

cmd/generator/main.go
lieut-data: Oh, I see what you mean now! Fixed -- sorry, I was thinking about a different kind of splitting. Or something. No idea now :)

lieut-data:
@marianunez / @hanzei, I've re-requested review just to get your eyes on the updates to support a `ReleaseNotesURL`. I decided to include it in this PR instead of spawning a new review cycle.

Right now, my `plugins.json` includes the updates from https://github.com/mattermost/mattermost-marketplace/pull/19, but if that gets merged, I'll just update here.

PR #21, opened 2019-11-13
MM-20135 - Update API to return single-signature for each plugin

ali-farooq0:
Pivoting to single-signature architecture as per discussion https://community-daily.mattermost.com/core/pl/ngzueo6drpgwdnrugkuct5nfrw


internal/model/plugin.go
```diff
@@ -22,7 +22,7 @@ type Plugin struct {
 	HomepageURL     string                    `json:"homepage_url"`
 	IconData        string                    `json:"icon_data"`
 	DownloadURL     string                    `json:"download_url"`
-	Signatures      []*PluginSignature        `json:"signatures"`
+	Signature       *PluginSignature          `json:"signature"`
```
ali-farooq0: `signature.signature` definitely stutters.

Had a few ideas but not sure which one to go with:
`signtuare_details.signature` or `signature.encoded_signature` or `signature.encoded`? What do you guys think?

iomodo: How about getting rid of `PluginSignature` struct?

ali-farooq0: Updated it to 

```
type Plugin struct {
	HomepageURL     string                    `json:"homepage_url"`
	IconData        string                    `json:"icon_data"`
	DownloadURL     string                    `json:"download_url"`
	ReleaseNotesURL string                    `json:"release_notes_url"`
	UpdatedAt       time.Time                 `json:"updated_at"`

	// Signature represents a signature of a plugin saved in base64 encoding.
	Signature string `json:"signature"`
	// SignaturePublicKeyHash represents an abbreviated public key fingerprint,
	// hashed using SHA-1 algorithm.
	SignaturePublicKeyHash string `json:"signature_public_key_hash"`
}
```

lieut-data: Looks like my review was out-of-date relative to these comments, but to pivot, maybe we just eliminate `SignaturePublicKeyHash` altogether?

lieut-data: I also note that in the GPG world, if we kept this, we may want to actually pivot to tracking the long key id instead of the hash of the file to be more idiomatic.

cmd/generator/main.go
```diff
@@ -318,25 +326,20 @@ func getFromTarFile(reader *tar.Reader, filepath string) ([]byte, error) {
...
-			return nil, errors.Wrap(err, "Can't get signature from the asset")
-		}
-
-		signature := &model.PluginSignature{
-			Signature:     sig,
-			PublicKeyHash: hash,
-		}
-		signatures = append(signatures, signature)
+func downloadSignature(asset *github.ReleaseAsset) (*model.PluginSignature, error) {
+	hash, err := getPublicKeyHashFromAsset(*asset)
```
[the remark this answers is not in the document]
ali-farooq0: ah nice! Had no idea you could extract out the full hash. Just confirming if we can do that through the API.


Alright tried it myself, you can't get the hash but you can use the `IssuerKeyId` and maybe that's good enough for now.

k will remove public hash.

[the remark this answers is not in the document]
ali-farooq0:
```
keyRingReader, err := os.Open("development-public-key.asc")
	if err != nil {
		fmt.Println(err)
		return
	}

	signature, err := os.Open("com.mattermost.demo-plugin-0.3.0.tar.gz-F3FACE.sig")
	if err != nil {
		fmt.Println(err)
		return
	}

	verificationTarget, err := os.Open("com.mattermost.demo-plugin-0.3.0.tar.gz")
	if err != nil {
		fmt.Println(err)
		return
	}

	keyring, err := openpgp.ReadArmoredKeyRing(keyRingReader)
	if err != nil {
		fmt.Println("Read Armored Key Ring: " + err.Error())
		return
	}

	p, err := packet.Read(signature)
	if err != nil {
		fmt.Println("PPP: " + err.Error())
		return
	}

	sig, ok := p.(*packet.Signature)
	if !ok {
		fmt.Println("!ok: ")
		return
	}

	keys := keyring.KeysById(*sig.IssuerKeyId)

	fmt.Printf("sig: %x\n", *sig.IssuerKeyId)
	fmt.Printf("keys: %+v\n", keys[0].PublicKey.KeyIdString())
```

internal/model/plugin.go
ali-farooq0: Yeah i made a small POC to confirm if `Signature.IssuerKeyId` would do the trick and looks like it'll work so we can get rid of `public_key_hash` completely.

lieut-data:
We're running into GitHub actions issues right now, but we've independently verified this. Also, heads up @prapti that I'm going to skip `2: QA Review` on this PR since we're effectively considering to be part of the feature branch, despite going to `master` in this case.

PR #22, opened 2019-11-14
Fixed wrapping a nil error

ali-farooq0:


ali-farooq0:
@lieut-data whenever you get a chance can you merge this as well?

PR #23, opened 2019-11-18
Update plugins.json

iomodo:
This PR will update signatures in plugins.json file


plugins.json
```diff
@@ -1,1165 +1,1165 @@
 [
-  {
```
ali-farooq0: can we continue to use 2-space spacing, otherwise its hard to see what changed here :(

ali-farooq0:
should probably also update `statik.go`

ali-farooq0:
Updated spacing and statik.go as @prapti needs to use this to test

PR #24, opened 2019-11-28
[MM-18809] Update signatures with production public key

iomodo:
This PR will update `plugins.json` file.

Fixes https://mattermost.atlassian.net/browse/MM-18809


plugins.json
```diff
@@ -526,7 +618,7 @@
       },
       "settings_schema": {
         "header": "",
-        "footer": "Use this webhook URL to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n`https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET&team=TEAMURL&channel=CHANNELURL`",
+        "footer": "Use this webhook URL to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n`https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET\u0026team=TEAMURL\u0026channel=CHANNELURL`",
```
ali-farooq0: @iomodo is this unicode `\u0026` encoding intentional? Is `jq` doing this?

plugins.json
```diff
@@ -1162,4 +1254,4 @@
     },
     "updated_at": "2019-10-17T16:37:46Z"
   }
-]
+]
```
ali-farooq0: empty line

plugins.json
iomodo: generator generated this way, not sure why

plugins.json
```diff
@@ -419,12 +419,104 @@
...
+    "download_url": "https://github.com/mattermost/mattermost-plugin-jira/releases/download/v2.3.0-rc1/jira-2.3.0.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-jira/releases/tag/v2.3.0-rc1",
```
lieut-data: I've pinged @DHaussermann and @levb to clarity here, since I'm wondering where there is a "released" rc1 version on GitHub.

plugins.json
lieut-data: Yeah, this worries me slightly too. It technically won't cause a problem, but good to get to the bottom of why.

ali-farooq0: Agreed. I normally do `go run ./cmd/generator --github-token $GITHUB_TOKEN --debug | jq > plugins.json` and don't notice the encoding locally 🤔

plugins.json
lieut-data: @iomodo, I've gone ahead and marked this version of the plugin as pre-release. Can you change the default `include-pre-release` flag in the `generator` to be false and then re-generate to exclude this release?

ali-farooq0:
Thanks @prapti, i think that should suffice.

PR #27, opened 2019-11-29
enable golangci-lint

lieut-data:
Enable golangci-lint in the new CircleCI configuration, fixing the reported issues along the way. I disabled any tests where it didn't make sense to fix the recommended changes.


.circleci/config.yml
```diff
@@ -19,21 +19,51 @@ commands:
   check-uncommitted:
     description: Check for uncommitted changes.
     steps:
-      - run: 
-          name: Check for uncomitted changes using git porcelain command.
+      - run:
+          name: Check for uncomitted changes.
```
lieut-data: Make this shorter in the CircleCI dashboard.

.golangci.yml
```diff
@@ -0,0 +1,49 @@
...
+linters:
+  disable-all: true
+  enable:
+    # - bodyclose
+    - deadcode
+    - depguard
+    - dogsled
+    - dupl
+    - errcheck
+    # - funlen
```
lieut-data: The `len` checks are artificial and annoying.

.golangci.yml
```diff
@@ -0,0 +1,49 @@
...
+    # - gochecknoinits
```
lieut-data: We use `init` for `statik`.

.golangci.yml
```diff
@@ -0,0 +1,49 @@
...
+    - goconst
+    - gocritic
+    - gocyclo
+    - gofmt
+    - goimports
+    - golint
+    # - gosec
```
lieut-data: `gosec` complains about our `http.Get` calls using URLs provided by the GitHub client. No way around this as far as I can see.

.golangci.yml
```diff
@@ -0,0 +1,49 @@
...
+    - gosimple
+    - govet
+    - ineffassign
+    - interfacer
+    # - lll
```
lieut-data: Line length limits are annoying.

Makefile
```diff
@@ -18,20 +18,18 @@ generate:
 	cp plugins.json data/static/
 	go generate ./...
 
-## Runs govet and gofmt against all packages.
+## Runs go vet and golangci-lint against all packages.
```
lieut-data: Keeping `go vet` around for sane errors when the code doesn't pass syntax tests. golangci-lint obscures this behind messages about being unable to run a particular checker.

Makefile
```diff
@@ -18,20 +18,18 @@ generate:
...
-govet:
+check-style:
 	go vet ./...
 
-## Runs lint against all packages.
-.PHONY: lint
-lint:
-	GO111MODULE=off go get -u golang.org/x/lint/golint
-	golint -set_exit_status ./...
+# https://stackoverflow.com/a/677212/1027058 (check if a command exists or not)
```
lieut-data: I do this here for developers, and in the CircleCI configuration for caching purposes.

cmd/generator/logger.go
```diff
@@ -9,17 +9,3 @@ var logger *log.Logger
 func init() {
 	logger = log.New()
 }
-
```
lieut-data: Unused code.

cmd/generator/main.go
```diff
@@ -274,7 +274,8 @@ func getReleasePlugin(release *github.RepositoryRelease, repository *github.Repo
 	logger.Debugf("found latest release %s", releaseName)
 
 	downloadURL := ""
-	var signatureAsset *github.ReleaseAsset
+	var signatureAsset github.ReleaseAsset
```
lieut-data: Avoid using references to for loop variables.

cmd/generator/main.go
```diff
@@ -325,13 +327,14 @@ func getReleasePlugin(release *github.RepositoryRelease, repository *github.Repo
 
 	// If no plugin in existing database or the updated timestamp has changed, attempt to download and inspect manifest.
 	if plugin == nil || updatedAt.IsZero() || plugin.UpdatedAt.Before(updatedAt) {
-		if plugin == nil {
+		switch {
```
lieut-data: Use switch instead.

cmd/generator/main.go
```diff
@@ -450,38 +453,16 @@ func getSignatureFromAsset(asset github.ReleaseAsset) (string, error) {
 	return base64.StdEncoding.EncodeToString(sigFile), nil
 }
 
-func getLatestRelease(ctx context.Context, client *github.Client, repoName string, includePreRelease bool) (*github.RepositoryRelease, error) {
```
lieut-data: Unused code.

cmd/generator/main.go
```diff
@@ -450,38 +453,16 @@ func getSignatureFromAsset(asset github.ReleaseAsset) (string, error) {
...
-
-	return latestRelease, nil
-}
-
 func getIcon(ctx context.Context, icon string) ([]byte, error) {
 	if strings.HasPrefix(icon, "http") {
 		logger.Debugf("fetching icon from url %s", icon)
 
-		resp, err := http.Get(icon)
+		req, err := http.NewRequestWithContext(ctx, "GET", icon, nil)
```
lieut-data: `ctx` was unused

internal/api/plugins_test.go
```diff
@@ -98,118 +98,118 @@ func TestPlugins(t *testing.T) {
 	})
 
 	t.Run("plugins", func(t *testing.T) {
-		plugin1_V1Min515 := &model.Plugin{
+		plugin1V1Min515 := &model.Plugin{
```
lieut-data: Underscores in variable names are now forbidden.

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+trim_trailing_whitespace = true
+
+[*.go]
+indent_style = tab
+
+[*.json]
+indent_size = 2
```
hanzei: I've noticed that we use an `indent_size` in https://github.com/mattermost/mattermost-plugin-starter-template/blob/master/.editorconfig#L16. Maybe we want to change this to be consistent with the starter template?

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+
+[Makefile,*.mk]
+indent_style = tab
+
+[*.md]
+indent_style = space
+indent_size = 4
```
hanzei: Would you please add 
```suggestion
indent_size = 4
trim_trailing_whitespace = false
```

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+
+[.golangci.yml]
```
hanzei: Maybe we merge this with the `*.json` rule?

.golangci.yml
```diff
@@ -0,0 +1,49 @@
+service:
+  golangci-lint-version: 1.21.0 # Keep this in sync with .circleci/config.yml and Jenkinsfile.pr
```
hanzei:
```suggestion
  golangci-lint-version: 1.21.0 # Keep this in sync with .circleci/config.yml
```

Makefile
```diff
@@ -18,20 +18,18 @@ generate:
...
-## Runs govet and gofmt against all packages.
+## Runs go vet and golangci-lint against all packages.
 .PHONY: check-style
-check-style: govet lint
-
-## Runs govet against all packages.
-.PHONY: govet
-govet:
+check-style:
 	go vet ./...
```
hanzei: Do we still need this?

Makefile
hanzei: Maybe we run `make test` and `make check-style` in parallel? This way `make test` will fail, if there is a syntax error.

> golangci-lint obscures this behind messages about being unable to run a particular checker.

I've not encountered this. Would you please provide an example of an error that gets hidden?

.circleci/config.yml
```diff
@@ -19,21 +19,51 @@ commands:
...
+        description: Prefix for cache key to store the binary.
+    steps:
+      - restore_cache:
+          name: Restore golangci-lint cache
+          keys: ['<< parameters.prefix >>-golangci-lint-{{ arch }}-<< parameters.version >>']
+      - run:
+          name: Install golangci-lint
+          command: |
+            mkdir -p << parameters.gobin >>
+            command -v << parameters.gobin >>/golangci-lint && exit
```
hanzei: Nice solution :+1:

.circleci/config.yml
```diff
@@ -19,21 +19,51 @@ commands:
...
+
+            download=https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh
+            wget -O- -q $download | sh -s -- -b << parameters.gobin >>/ v<< parameters.version >>
```
hanzei: I would vote for not running this completely silenced.
```suggestion
            wget -nv- -q $download | sh -s -- -b << parameters.gobin >>/ v<< parameters.version >>
```

hanzei: Nit: The server uses `curl` instead (https://github.com/mattermost/mattermost-server/blob/master/.circleci/config.yml#L67). Maybe we want to sync this?

.circleci/config.yml
```diff
@@ -19,21 +19,51 @@ commands:
...
+        default: 1.21.0
+      gobin:
+        type: string
+        default: /go/bin
+      prefix:
+        type: string
+        default: v1
+        description: Prefix for cache key to store the binary.
+    steps:
+      - restore_cache:
```
hanzei: I really like the caching

.golangci.yml
```diff
@@ -0,0 +1,49 @@
...
+  gofmt:
+    simplify: true
+  govet:
+    check-shadowing: true
+    enable-all: true
+
+linters:
+  disable-all: true
+  enable:
+    # - bodyclose
```
hanzei: We can enable this, if we drop the check at https://github.com/mattermost/mattermost-marketplace/blob/master/internal/api/client.go#L29, which seams reasonable to me.


Other question: Why are making sure to read the full body in https://github.com/mattermost/mattermost-marketplace/blob/master/internal/api/client.go#L30? I wasn't able to find a reason for this.

.golangci.yml
```diff
@@ -0,0 +1,49 @@
...
+    - deadcode
+    - depguard
```
hanzei: You didn't configure a list for `depguard`. Without a configured list, the linter will do nothing. Did you intent to blacklist any dependencies?

.golangci.yml
hanzei: I would also say the same for `dupl`, `dogsled`, `gocyclo`. But we can remove them, if we run into false positives.

.golangci.yml
hanzei: We can add a exclude rule for this in the code using `//nolint:gosec` (https://github.com/golangci/golangci-lint#nolint)

.golangci.yml
```diff
@@ -0,0 +1,49 @@
...
+  timeout: 5m
+  modules-download-mode: readonly
+
+linters-settings:
+  gofmt:
+    simplify: true
+  govet:
+    check-shadowing: true
+    enable-all: true
+
```
hanzei: Let's converge on one local 
```suggestion
  misspell:
    locale: US

```

hanzei: I usually set `golint` to a more aggressive setting and have not encountered any false positives yet.

```suggestion
  golint:
    min-confidence: 0

```

hanzei: I like the option for `goimports`, that forces you to split the imports into three parts. This would also be consistent with the Mattermost server.

```suggestion
  goimports:
    local-prefixes: github.com/mattermost/mattermost-marketplace

```

hanzei: Maybe we want to be more aggressive with `goconst`:
```suggestion
  goconst:
    min-len: 2
    min-occurrences: 2

```

.golangci.yml
lieut-data: Ugh... I'm all for extrinsic tests, but when a linter asserts something so wildly limiting that we have to start tuning it with comments eslint-style, I feel like we've lost the spirit of what's idiomatic Go.

.editorconfig
lieut-data: Happy to do so -- why is this required?

.editorconfig
lieut-data: With the change to `indent_size = 4` above, I think we can leave it separate now?

Makefile
lieut-data: With it, I get a nice clean error message on a syntax error:
```
> make check-style
go vet ./...
# github.com/mattermost/mattermost-marketplace/cmd/marketplace
vet: cmd/marketplace/main.go:18:30: missing ',' before newline in argument list (and 6 more errors)
make: *** [check-style] Error 2
```

without it ... not so much:
```
make check-style
golangci-lint run ./...
WARN [runner] Can't run linter goanalysis_metalinter: SA5007: failed prerequisites: buildssa@github.com/mattermost/mattermost-marketplace/cmd/marketplace
WARN [runner] Can't run linter unused: buildssa: analysis skipped: errors in package: [/Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:18:30: missing ',' before newline in argument list /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:19:1: expected operand, found '}' /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:22:2: missing ',' in argument list /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:23:48: missing ',' before newline in composite literal /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:24:13: missing ',' before newline in composite literal /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:25:3: missing ',' before newline in argument list /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:26:1: expected operand, found '}' /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:18:30: missing ',' before newline in argument list /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:19:1: expected operand, found '}' /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:22:2: missing ',' in argument list /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:23:48: missing ',' before newline in composite literal /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:24:13: missing ',' before newline in composite literal /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:25:3: missing ',' before newline in argument list /Users/jesse/Documents/mattermost/src/marketplace/cmd/marketplace/main.go:26:1: expected operand, found '}']
```

.circleci/config.yml
lieut-data: Looks like the Mattermost server already silences the `curl` -- any specific concern here?

lieut-data: I'd be inclined to just leave as-is -- this is copied and pasted from an existing orb, and I think it's probably fine to diverge.

.circleci/config.yml
lieut-data: Me too! I note also that /not/ having multiple CircleCI steps saves us the cost of restoring the cache each time.

.golangci.yml
lieut-data: I was under the impression the `nil` check was necessary in some cases. I don't mind the linting improvements, but I'm hesitant to start to force it into awkward situations like this.

The reason to consume the whole body in general is to enable keep-alive. It may not matter a lot in our particular case, but there doesn't seem to be much harm in adopting this pattern as-is from the mattermost server.

.golangci.yml
lieut-data: Ah, good catch. I literally just enabled everything and fixed what I could. I didn't realize depguard required further configuration. Will remove.

.golangci.yml
hanzei: Hm, I see that comments are not the right way.

Maybe we just ignore this specific rule of `gosec` using
```yml
issues:
  exclude:
    - "G107: Potential HTTP request made with variable url"
```
?

.editorconfig
hanzei: You can create line breaks in markdown with a double white space at the end of the line. `trim_trailing_whitespace = true` remove them, which might be unexpecting.

.editorconfig
hanzei: 0/5, it can stay

.circleci/config.yml
hanzei: No specific ones. I seams like one of the cases where if something does wrong, it want to see the error message. 1/5

Makefile
```diff
@@ -18,24 +18,22 @@ generate:
...
-lint:
-	GO111MODULE=off go get -u golang.org/x/lint/golint
-	golint -set_exit_status ./...
+# https://stackoverflow.com/a/677212/1027058 (check if a command exists or not)
+	@if ! [ -x "$$(command -v golangci-lint)" ]; then \
+		echo "golangci-lint is not installed. Please see https://github.com/golangci/golangci-lint#install for installation instructions."; \
+		exit 1; \
+	fi; \
+
+	golangci-lint run --max-same-issues=10 ./...
```
hanzei: Mind setting this in the config file?

Makefile
```diff
@@ -18,24 +18,22 @@ generate:
...
 
 ## Runs test against all packages.
 .PHONY: test
-test:
+test: check-style
```
hanzei: In all other projects we don't run style checks alongside with tests. My concern is, that the style checks might take some seconds and prevent devs from quickly running the tests. Also, I think it's fine to run tests without passing style checks.

Makefile
hanzei: I was ambiguous here, sorry. I was talking about running them in parallel in circleci, not in the Makefile.

lieut-data: Apologies for misunderstanding -- what's the benefit to running them in parallel on CircleCI? I don't want my tests to fail because of syntax errors /and/ my lint to fail for the same reason. 

Plus, at least for a repository of this size, running them serially is "faster" since we only need to wait for a single CircleCI process, that only has to restore/save cache once.

Makefile
lieut-data: Seems reasonable: will break it back out.

Makefile
hanzei: Good point, I'm over engineering.

.golangci.yml
lieut-data: Sounds good!

agnivade:
Just wanted to quickly echo @hanzei's comment on the separate circle stage for installing golangci-lint. It keeps things very clean. We should definitely do this on all repos. :+1:

PR #31, opened 2019-12-18
[MM-21123] Add official label to all plugins

hanzei:
#### Summary
This PR add the "official" label to all plugins served by the marketplace, given that there are no community plugins at the moment. It also adds a `/api/v1/labels` that returns all labels, which is used for debugging.

I also added some documentation for the health check endpoint and fixed inconsistencies.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21123

#### Related Pull Requests
- Has Server changes: https://github.com/mattermost/mattermost-server/pull/13425


internal/model/label.go
```diff
@@ -0,0 +1,14 @@
...
+type Label struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	URL         string `json:"url"`
+}
+
+var OfficialLabel Label = Label{
+	Name:        "Official",
+	Description: "This plugin is maintained by Mattermost",
+	URL:         "https://mattermost.com/pl/default-community-plugins",
```
hanzei: This will need to point to the docs. We need to set up the link before the 5.20 release.

internal/model/label.go
```diff
@@ -0,0 +1,14 @@
...
+// Label represents a label shown in the Marketplace UI.
+type Label struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	URL         string `json:"url"`
+}
+
+var OfficialLabel Label = Label{
+	Name:        "Official",
+	Description: "This plugin is maintained by Mattermost",
```
hanzei: @aaronrothschild Just want to double check that you are fine with this wording.

aaronrothschild: Yes, I think that's accurate.

internal/api/health_check.go
```diff
@@ -20,15 +20,18 @@ type healthCheckResponse struct {
 	Description string                       `json:"description"`
 }
 
+// initHealthCheck health check endpoints on the given router.
```
lieut-data: Small GoDoc-tweak, to create a sentence:
```suggestion
// initHealthCheck initializes health check endpoints on the given router.
```

internal/api/health_check.go
```diff
@@ -20,15 +20,18 @@ type healthCheckResponse struct {
...
 	addContext := func(handler contextHandlerFunc) *contextHandler {
 		return newContextHandler(context, handler)
 	}
 
 	pluginsRouter := apiRouter.PathPrefix("/health").Subrouter()
-	pluginsRouter.Handle("", addContext(handleHealthCheck)).Methods("GET")
+	pluginsRouter.Handle("", addContext(handleHealthCheck)).Methods(http.MethodGet)
 }
 
+// handleHealthCheck responds to GET /api/v1/health,
```
lieut-data: nit: was the line wrap here strictly necessary?

internal/model/label.go
lieut-data: Should we trail with a period, or do we not do that with the labels today?

internal/model/plugin.go
```diff
@@ -8,16 +8,16 @@ import (
 	mattermostModel "github.com/mattermost/mattermost-server/model"
 )
 
-// Plugin represents a Mattermost plugin in the marketplace.
+// Plugin represents a Mattermost plugin in the Marketplace.
```
lieut-data: nit: I've been aiming to capitalize only when in the proper noun context: `Plugin Marketplace`. Should we do that here too if we want to capitalize at all?
```suggestion
// Plugin represents a Mattermost plugin in the Plugin Marketplace.
```

internal/model/plugin.go
```diff
@@ -8,16 +8,16 @@ import (
...
-	// Signature represents a signature of a plugin saved in base64 encoding.
-	Signature string                    `json:"signature"`
-	Manifest  *mattermostModel.Manifest `json:"manifest"`
-	UpdatedAt time.Time                 `json:"updated_at"`
+	HomepageURL     string                    `json:"homepage_url"`
+	IconData        string                    `json:"icon_data"`
+	DownloadURL     string                    `json:"download_url"`
+	ReleaseNotesURL string                    `json:"release_notes_url"`
+	Labels          []Label                   `json:"labels"`
+	Signature       string                    `json:"signature"` // Signature represents a signature of a plugin saved in base64 encoding.
```
lieut-data: To confirm, will GoDoc pick up on inline comments here? I'd also be open to either just removing this comment, or expanding to all properties.

internal/api/health_check.go
hanzei: Without the wrapping the line has 130 characters, which is too much IMO if it can be avoided.	But I'm 0/5, because we don't enforce line length.

internal/model/plugin.go
hanzei: Good catch. Will do :+1:

internal/model/plugin.go
hanzei: Yes, godoc picks up inline comments.

I'm all for more documentation, so I will add some comments.

hanzei: Yes, godoc picks up inline comments.

I'm all for more documentation, but I'm would only document the ones that are not self explaining.

internal/model/label.go
hanzei: I'm 0/5.

hanzei:
Will fix the merge convict once dev review is done

internal/model/label.go
hanzei: @aaronrothschild What do you think about the Jesse's proposal?

hanzei:
@lieut-data @prapti I did add one more change in https://github.com/mattermost/mattermost-marketplace/pull/31/commits/9bd118832d75318dab124c52273ed0098e4699b8 after your review. Given how small the change is, I'm not re-requesting your reviews. Still, fell free to leave comments.

internal/model/label.go
aaronrothschild: no trailing dot is good

aaronrothschild:
@hanzei Can I get a screenshot of what it looks like now?

internal/model/label.go
hanzei: :+1: I changed this also for the `Local` label in https://github.com/mattermost/mattermost-server/pull/13425/commits/018fd8d6961f0fee06df69f8aa394b0227305d83

hanzei:
@lieut-data Good question. It kind of works in most cases. I've updated the screenshots in the webapp PR. Please see https://github.com/mattermost/mattermost-webapp/pull/4547.

@aaronrothschild Please also see https://github.com/mattermost/mattermost-webapp/pull/4547

PR #33, opened 2019-12-19
Update URL in README

hanzei:
#### Summary
Fix URL in README

#### Ticket Link
None


lieut-data:
@prapti, this is a good candidate for skipping QA testing, but just sticking with the current process until we publish our new strategy with the team early next year :)

PR #35, opened 2020-01-15
Add generator add command

hanzei:
#### Summary
This PR add a `generator add` command, which is used to add community plugins to the Marketplace Database.

To update the database you would e.g. run `go run ./cmd/generator/ add matterpoll v1.5.0 | jq | sponge plugins.json`.

I did a lot of refactoring in `cmd/generator/main.go` to keep the code DRY. I've rebuild the database from scratch and there was no difference to the existing one. Hence, I'm confident, I didn't add a regression.

I've tested the changes by downloading the release from https://github.com/hanzei/plugin-test/releases/tag/v1.4.0. The generated changed to the database can be found below.

`release_notes_url` is not set, because there is not mechanism in the manifest to define it and we can't rely on the GitHub URL, as the repo might not be located at GitHub.

`signature` is not set, because there is not signature for that release. I'm working with @cpanato to test this command with a plugin that also has a signature.

<details>
<summary>Details</summary>
```json
 {
    "homepage_url": "https://github.com/matterpoll/matterpoll",
    "icon_data": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz4KPHN2ZyB3aWR0aD0iNjFweCIgaGVpZ2h0PSI2MXB4IiB2aWV3Qm94PSIwIDAgNjEgNjEiIHZlcnNpb249IjEuMSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayI+CiAgICA8IS0tIEdlbmVyYXRvcjogU2tldGNoIDU5ICg4NjEyNykgLSBodHRwczovL3NrZXRjaC5jb20gLS0+CiAgICA8dGl0bGU+TG9nbyBEYXJrPC90aXRsZT4KICAgIDxkZXNjPkNyZWF0ZWQgd2l0aCBTa2V0Y2guPC9kZXNjPgogICAgPGcgaWQ9IlBhZ2UtMSIgc3Ryb2tlPSJub25lIiBzdHJva2Utd2lkdGg9IjEiIGZpbGw9Im5vbmUiIGZpbGwtcnVsZT0iZXZlbm9kZCI+CiAgICAgICAgPGcgaWQ9IkNvbG9yZWQtLS0iIHRyYW5zZm9ybT0idHJhbnNsYXRlKC0xODYuMDAwMDAwLCAtMjUwLjAwMDAwMCkiIGZpbGw9IiMwMDAwMDAiPgogICAgICAgICAgICA8ZyBpZD0iTG9nby1EYXJrIiB0cmFuc2Zvcm09InRyYW5zbGF0ZSgxODYuMDAwMDAwLCAyNTAuMDAwMDAwKSI+CiAgICAgICAgICAgICAgICA8cGF0aCBkPSJNMjQuNzY5NzQ2MywwLjAwMDEzNTYwMjk3NyBMMjQuNzcwMzczNiwxMS43Njk4Mjc4IEMxNy4wMzk5MjU3LDE0LjE5Nzg2MjMgMTEuNDMyNTIwNSwyMS40MTk5MzkxIDExLjQzMjUyMDUsMjkuOTUxNjk1NSBDMTEuNDMyNTIwNSwzOC40ODM0NTE5IDE3LjAzOTkyNTcsNDUuNzA1NTI4NyAyNC43NzAzNzM2LDQ4LjEzMzU2MzIgTDI0Ljc3LDYwLjQzOCBMMiw2MC40Mzg0MTY5IEMwLjg5NTQzMDUsNjAuNDM4NDE2OSAtMS40MjEwODU0N2UtMTQsNTkuNTQyOTg2NCAtMS4wNjU4MTQxZS0xNCw1OC40Mzg0MTY5IEwwLDI4LjA0NjI3NTQgTDAuMDU4NTkxOTg4MywyOC4wNDYwMjI2IEMwLjkyMzYxNDMyNCwxNC4wMjA1NTA1IDExLjI3MTYxMTMsMi41NjA5NzU5NyAyNC43Njk3NDYzLDAuMDAwMTM1NjAyOTc3IFogTTYwLjgxNDQ4NzgsMjYuODE5MzI3MyBDNjAuOTE5NTkzNywyNy44NDkxNDQ3IDYwLjk3MzQ0MjgsMjguODk0MTMxNSA2MC45NzM0NDI4LDI5Ljk1MTY5NTUgQzYwLjk3MzQ0MjgsNDQuODM1MzUxNiA1MC4zMDc4NjY3LDU3LjIyNzkyOSAzNi4yMDI2OTcxLDU5LjkwMzQ0NSBMMzYuMjAzMDY5MSw0OC4xMzM1NjMyIEM0My45MzM1MTcsNDUuNzA1NTI4NyA0OS41NDA5MjIyLDM4LjQ4MzQ1MTkgNDkuNTQwOTIyMiwyOS45NTE2OTU1IEw0OS41MzksMjkuODQgWiBNMzYuMjAyNjk3MSwtNS4zOTc3OTcwOWUtMDUgQzQ1LjQ5MTA4NzksMS43NjE3OTkyNCA1My4yODc5NTA2LDcuNzM3MzEwNDQgNTcuNTI5ODg1MSwxNS44NjMwNzk3IEw0Ni4wNDA1MDUzLDE4Ljk0MjU1NSBDNDMuNjU5MTAwMywxNS41ODQxOTM3IDQwLjIxNzMyNCwxMy4wMzA2NTM4IDM2LjIwMzA2OTEsMTEuNzY5ODI3OCBaIiBpZD0iQ29tYmluZWQtU2hhcGUiPjwvcGF0aD4KICAgICAgICAgICAgPC9nPgogICAgICAgIDwvZz4KICAgIDwvZz4KPC9zdmc+",
    "download_url": "https://github.com/hanzei/plugin-test/releases/download/v1.4.0/com.github.matterpoll.matterpoll-1.4.0.tar.gz",
    "release_notes_url": "",
    "labels": null,
    "signature": "",
    "manifest": {
      "id": "com.github.matterpoll.matterpoll",
      "name": "Matterpoll",
      "description": "Polling feature for Mattermost's custom slash command",
      "homepage_url": "https://github.com/matterpoll/matterpoll",
      "support_url": "https://github.com/matterpoll/matterpoll/issues",
      "icon_path": "assets/logo_dark.svg",
      "version": "1.4.0",
      "min_server_version": "5.14.0",
      "server": {
        "executables": {
          "linux-amd64": "server/dist/plugin-linux-amd64",
          "darwin-amd64": "server/dist/plugin-darwin-amd64",
          "windows-amd64": "server/dist/plugin-windows-amd64.exe"
        },
        "executable": ""
      },
      "webapp": {
        "bundle_path": "webapp/dist/main.js"
      },
      "settings_schema": {
        "header": "",
        "footer": "* To report an issue, make a suggestion or a contribution, [check the repository](https://github.com/matterpoll/matterpoll).",
        "settings": [
          {
            "key": "Trigger",
            "display_name": "Trigger Word",
            "type": "text",
            "help_text": "Trigger Word must be unique, and cannot begin with a slash or contain any spaces.",
            "placeholder": "",
            "default": "poll"
          },
          {
            "key": "ExperimentalUI",
            "display_name": "Experimental UI",
            "type": "bool",
            "help_text": "When true, Matterpoll will render poll posts with a rich UI. The rich UI is not available on mobile app.",
            "placeholder": "",
            "default": false
          }
        ]
      }
    },
    "updated_at": "2020-01-15T12:26:29.104271951Z"
 }
```
</details>

#### Open Questions
- It it worth to add a make target e.g. `make add matterpoll v1.5.0`?
- With these changes it's not longer possible to remove the database and rebuild it from scratch. Are there situations where is behaviours is needed or wanted?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21747

### DIFF
--- a/cmd/generator/add.go
+++ b/cmd/generator/add.go
@@ -23,8 +23,8 @@ var addCmd = &cobra.Command{
 	Use:   "add [repo] [tag]",
 	Short: "Add a plugin release to the plugins.json database",
 	Long: "The generator commands allows adding a specific plugin release to the database by using this command.\n\n" +
-		"The release has to be build first using the cut plugin command, which also uploads it to https://plugins-store.test.mattermost.com/release/. " +
-		"This localtion is used to fetch the plugin release.",
+		"The release has to be built first using the /mb cutPlugin command, which also uploads it to https://plugins-store.test.mattermost.com/release/. " +
+		"This location is used to fetch the plugin release.",
 	Example: `  generator add matterpoll v1.5.1`,
 	Args:    cobra.ExactArgs(2),
 	RunE: func(command *cobra.Command, args []string) error {
--- a/cmd/generator/add.go
+++ b/cmd/generator/add.go
@@ -20,8 +20,11 @@ func init() {
 }
 
 var addCmd = &cobra.Command{
-	Use:     "add [repo] [tag]",
-	Short:   "Add a plugin release to the plugins.json database",
+	Use:   "add [repo] [tag]",
+	Short: "Add a plugin release to the plugins.json database",
+	Long: "The generator commands allows adding a specific plugin release to the database by using this command.\n\n" +
+		"The release has to be build first using the cut plugin command, which also uploads it to https://plugins-store.test.mattermost.com/release/. " +
+		"This localtion is used to fetch the plugin release.",
 	Example: `  generator add matterpoll v1.5.1`,
 	Args:    cobra.ExactArgs(2),
 	RunE: func(command *cobra.Command, args []string) error {
--- a/cmd/generator/add.go
+++ b/cmd/generator/add.go
@@ -1,0 +1,92 @@
+package main
+
+import (
+	"archive/tar"
+	"bytes"
+	"encoding/json"
+	"os"
+	"regexp"
+	"time"
+
+	mattermostModel "github.com/mattermost/mattermost-server/model"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/mattermost/mattermost-marketplace/internal/model"
+)
+
+func init() {
+	generatorCmd.AddCommand(addCmd)
+}
+
+var addCmd = &cobra.Command{
+	Use:     "add [repo] [tag]",
+	Short:   "Add a plugin release to the plugins.json database",
+	Example: `  generator add matterpoll v1.5.1`,
+	Args:    cobra.ExactArgs(2),
+	RunE: func(command *cobra.Command, args []string) error {
+		command.SilenceUsage = true
+
+		plugins, err := InitCommand(command)
+		if err != nil {
+			return err
+		}
+
+		repo := args[0]
+		tag := args[1]
+		if !regexp.MustCompile("v[0-9]+.[0-9]+.[0-9]+$").MatchString(tag) {
+			return errors.Errorf("bad tag. Maybe a typo? You set this tag %s but we expected something like v2.3.4", tag)
+		}
+
+		bundleURL := "https://plugins-store.test.mattermost.com/release/" + repo + "-" + tag + ".tar.gz"
+		signatureURL := bundleURL + ".sig"
+
+		bundleData, err := downloadBundleData(bundleURL)
+		if err != nil {
+			return errors.Wrapf(err, "failed download bundle data")
+		}
+
+		manifestData, err := getFromTarFile(tar.NewReader(bytes.NewReader(bundleData)), "plugin.json")
+		if err != nil {
+			return errors.Wrap(err, "failed to read manifest from plugin bundle for release")
+		}
+
+		manifest := mattermostModel.ManifestFromJson(bytes.NewReader(manifestData))
+		if manifest == nil {
+			return errors.New("manifest nil after reading from plugin bundle for release")
+		}
+
+		var iconData string
+		if manifest.IconPath != "" {
+			iconData, err = getIconDataFromTarFile(bundleData, manifest.IconPath)
+			if err != nil {
+				return errors.Wrap(err, "failed to set icon")
+			}
+		}
+
+		signature, err := downloadSignature(signatureURL)
+		if err != nil {
+			return errors.Wrap(err, "failed to download plugin signature")
+		}
+
+		plugin := &model.Plugin{
+			HomepageURL:     manifest.HomepageURL,
+			IconData:        iconData,
+			DownloadURL:     bundleURL,
+			ReleaseNotesURL: "", // Not jet supported
+			Labels:          nil,
+			Signature:       signature,
+			Manifest:        manifest,
+			UpdatedAt:       time.Now().In(time.UTC),
+		}
+		plugin.Signature = ""
+
+		plugins = append(plugins, plugin)
+		err = json.NewEncoder(os.Stdout).Encode(plugins)
+		if err != nil {
+			return errors.Wrap(err, "failed to encode plugins result")
+		}
+
+		return nil
+	},
+}
--- a/cmd/generator/add.go
+++ b/cmd/generator/add.go
@@ -47,7 +47,7 @@ var addCmd = &cobra.Command{
 
 		bundleData, err := downloadBundleData(bundleURL)
 		if err != nil {
-			return errors.Wrapf(err, "failed download bundle data")
+			return errors.Wrapf(err, "failed downloading bundle data")
 		}
 
 		manifestData, err := getFromTarFile(tar.NewReader(bytes.NewReader(bundleData)), "plugin.json")
@@ -64,7 +64,7 @@ var addCmd = &cobra.Command{
 		if manifest.IconPath != "" {
 			iconData, err = getIconDataFromTarFile(bundleData, manifest.IconPath)
 			if err != nil {
-				return errors.Wrap(err, "failed to set icon")
+				return errors.Wrap(err, "failed to get icon")
 			}
 		}
 
@@ -83,7 +83,6 @@ var addCmd = &cobra.Command{
 			Manifest:        manifest,
 			UpdatedAt:       time.Now().In(time.UTC),
 		}
-		plugin.Signature = ""
 
 		plugins = append(plugins, plugin)
 		err = json.NewEncoder(os.Stdout).Encode(plugins)
--- a/cmd/generator/add.go
+++ b/cmd/generator/add.go
@@ -5,9 +5,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"os"
-	"regexp"
 	"time"
 
+	"github.com/blang/semver"
 	mattermostModel "github.com/mattermost/mattermost-server/model"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -34,8 +34,9 @@ var addCmd = &cobra.Command{
 
 		repo := args[0]
 		tag := args[1]
-		if !regexp.MustCompile("v[0-9]+.[0-9]+.[0-9]+$").MatchString(tag) {
-			return errors.Errorf("bad tag. Maybe a typo? You set this tag %s but we expected something like v2.3.4", tag)
+
+		if _, err = semver.ParseTolerant(tag); err != nil {
+			return errors.Wrapf(err, "%v is an invalid tag. Something like v2.3.4 is expected", tag)
 		}
 
 		bundleURL := "https://plugins-store.test.mattermost.com/release/" + repo + "-" + tag + ".tar.gz"
--- a/cmd/generator/main.go
+++ b/cmd/generator/main.go
@@ -116,7 +116,7 @@ var generatorCmd = &cobra.Command{
 				if len(plugin.IconData) == 0 {
 					if iconPath, ok := iconPaths[repositoryName]; ok {
 						var iconData string
-						iconData, err = getIconDataFromPath(ctx, iconPath)
+						iconData, err = getIconDataFromPath(iconPath)
 						if err != nil {
 							return errors.Wrapf(err, "failed to fetch icon for repository %s", repositoryName)
 						}
@@ -314,10 +314,12 @@ func getReleasePlugin(release *github.RepositoryRelease, repository *github.Repo
 		}
 
 		if plugin.Manifest.IconPath != "" {
-			plugin.IconData, err = getIconDataFromTarFile(bundleData, plugin.Manifest.IconPath)
+			var iconData string
+			iconData, err = getIconDataFromTarFile(bundleData, plugin.Manifest.IconPath)
 			if err != nil {
 				return nil, errors.Wrapf(err, "failed to set icon for release %s", releaseName)
 			}
+			plugin.IconData = iconData
 		}
 	} else {
 		logger.Debugf("skipping download since found existing plugin")
@@ -380,25 +382,25 @@ func downloadSignature(url string) (string, error) {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return "", errors.Errorf("received %d status code while downloading plugin bundle", resp.StatusCode)
+		return "", errors.Errorf("received %d status code while downloading plugin bundle from %v", resp.StatusCode, url)
 	}
 
-	sigFile, err := ioutil.ReadAll(resp.Body)
+	signature, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return "", errors.Wrapf(err, "failed to open downloaded signature file from %s", url)
+		return "", errors.Wrapf(err, "failed to read signature from %s", url)
 	}
-	return base64.StdEncoding.EncodeToString(sigFile), nil
+	return base64.StdEncoding.EncodeToString(signature), nil
 }
 
 func downloadBundleData(url string) ([]byte, error) {
 	resp, err := http.Get(url)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to download plugin bundle")
+		return nil, errors.Wrapf(err, "failed to download plugin bundle from %v", url)
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, errors.Errorf("received %d status code while downloading plugin bundle", resp.StatusCode)
+		return nil, errors.Errorf("received %d status code while downloading plugin bundle from %v", resp.StatusCode, url)
 	}
 
 	gzBundleReader, err := gzip.NewReader(resp.Body)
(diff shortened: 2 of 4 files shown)


cmd/generator/main.go
```diff
@@ -31,10 +31,11 @@ import (
 )
 
 func init() {
-	generatorCmd.PersistentFlags().String("github-token", "", "The optional GitHub token for API requests.")
 	generatorCmd.PersistentFlags().Bool("debug", false, "Whether to output debug logs.")
-	generatorCmd.PersistentFlags().Bool("include-pre-release", false, "Whether to include pre-release versions.")
-	generatorCmd.PersistentFlags().String("existing", "", "An existing plugins.json to help streamline incremental updates.")
+	generatorCmd.PersistentFlags().String("existing", "plugins.json", "An existing plugins.json to help streamline incremental updates.")
```
hanzei: Building the whole database from scratch doesn't make sense in most cases. Hence, it now opt in.

cmd/generator/add.go
```diff
@@ -0,0 +1,92 @@
...
+			ReleaseNotesURL: "", // Not jet supported
+			Labels:          nil,
+			Signature:       signature,
+			Manifest:        manifest,
+			UpdatedAt:       time.Now().In(time.UTC),
+		}
+		plugin.Signature = ""
+
+		plugins = append(plugins, plugin)
+		err = json.NewEncoder(os.Stdout).Encode(plugins)
```
hanzei: Out of scope of the PR, but I would like to get rid of the extra commands needed up update the file and only have to run `go run ./cmd/generator add matterpoll v1.5.1`

cmd/generator/add.go
```diff
@@ -0,0 +1,92 @@
...
+			HomepageURL:     manifest.HomepageURL,
+			IconData:        iconData,
+			DownloadURL:     bundleURL,
+			ReleaseNotesURL: "", // Not jet supported
+			Labels:          nil,
+			Signature:       signature,
+			Manifest:        manifest,
+			UpdatedAt:       time.Now().In(time.UTC),
+		}
+		plugin.Signature = ""
```
lieut-data: Is this intentional? (I know signatures are pending, but presumably it's already empty above?)

cmd/generator/main.go
lieut-data: Related to eliminating `jq` and `sponge`, I wonder if we could change this to be just a `database` flag, with the expectation of reading and writing directly to that file. It could still default to `plugins.json`.

cmd/generator/main.go
```diff
@@ -120,28 +106,21 @@ var generatorCmd = &cobra.Command{
 		for _, repositoryName := range repositoryNames {
 			logger.Debugf("querying repository %s", repositoryName)
 
-			releasePlugins, err := getReleasePlugins(ctx, client, repositoryName, includePreRelease, existingPlugins)
+			var releasePlugins []*model.Plugin
```
lieut-data: Can you elaborate on why this is now explicitly declared?

cmd/generator/main.go
```diff
@@ -120,28 +106,21 @@ var generatorCmd = &cobra.Command{
...
+			releasePlugins, err = getReleasePlugins(ctx, client, repositoryName, includePreRelease, existingPlugins)
 			if err != nil {
 				return errors.Wrapf(err, "failed to release plugin for repository %s", repositoryName)
 			}
 
 			for _, plugin := range releasePlugins {
 				if len(plugin.IconData) == 0 {
 					if iconPath, ok := iconPaths[repositoryName]; ok {
-						icon, err := getIcon(ctx, iconPath)
+						var iconData string
```
lieut-data: Similarly here?

cmd/generator/main.go
```diff
@@ -347,13 +314,10 @@ func getReleasePlugin(release *github.RepositoryRelease, repository *github.Repo
 		}
 
 		if plugin.Manifest.IconPath != "" {
-			iconData, err := getFromTarFile(tar.NewReader(bytes.NewReader(bundleData)), plugin.Manifest.IconPath)
+			plugin.IconData, err = getIconDataFromTarFile(bundleData, plugin.Manifest.IconPath)
```
lieut-data: This will end up overwriting `plugin.IconData` if an error occurs -- that /seems/ unexpected from a side effects perspective, IMO.

cmd/generator/main.go
```diff
@@ -406,55 +370,131 @@ func getFromTarFile(reader *tar.Reader, filepath string) ([]byte, error) {
...
-	if strings.HasPrefix(icon, "http") {
-		logger.Debugf("fetching icon from url %s", icon)
+func getIconDataFromTarFile(file []byte, path string) (string, error) {
+	iconData, err := getFromTarFile(tar.NewReader(bytes.NewReader(file)), path)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to read icon data from plugin bundle for path %s", path)
+	}
 
-		req, err := http.NewRequestWithContext(ctx, http.MethodGet, icon, nil)
+	logger.Debugf("using icon specified in manifest as %s", path)
```
lieut-data: nit: this log assumes it's being used in the context of a `manifest`, and that it's being "used" -- but the method is just reading an icon from the given tar file, and if reused, might not have the same side effects. Perhaps this log belongs at the caller instead?

cmd/generator/main.go
```diff
@@ -406,55 +370,131 @@ func getFromTarFile(reader *tar.Reader, filepath string) ([]byte, error) {
...
-	return signature, nil
-}
+	if resp.StatusCode != http.StatusOK {
+		return "", errors.Errorf("received %d status code while downloading plugin bundle", resp.StatusCode)
+	}
 
-func getSignatureFromAsset(asset github.ReleaseAsset) (string, error) {
-	url := asset.GetBrowserDownloadURL()
-	logger.Debugf("fetching signature file from %s", url)
+	sigFile, err := ioutil.ReadAll(resp.Body)
```
lieut-data: nit: call this `signature`? It's not a file yet, per se.

cmd/generator/main.go
```diff
@@ -406,55 +370,131 @@ func getFromTarFile(reader *tar.Reader, filepath string) ([]byte, error) {
...
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to open downloaded signature file from %s", url)
```
lieut-data:
```suggestion
		return "", errors.Wrapf(err, "failed to read signature from %s", url)
```

cmd/generator/main.go
```diff
@@ -406,55 +370,131 @@ func getFromTarFile(reader *tar.Reader, filepath string) ([]byte, error) {
...
 
-func downloadSignature(asset github.ReleaseAsset) (string, error) {
-	signature, err := getSignatureFromAsset(asset)
+func downloadSignature(url string) (string, error) {
+	logger.Debugf("fetching signature file from %s", url)
+
+	resp, err := http.Get(url)
 	if err != nil {
-		return "", errors.Wrap(err, "Can't get signature from the asset")
+		return "", errors.Wrapf(err, "failed to download signature file from %s", url)
```
lieut-data:
```suggestion
		return "", errors.Wrapf(err, "failed to download signature from %s", url)
```

cmd/generator/main.go
```diff
@@ -406,55 +370,131 @@ func getFromTarFile(reader *tar.Reader, filepath string) ([]byte, error) {
...
+	}
 
-		req, err := http.NewRequestWithContext(ctx, http.MethodGet, icon, nil)
+	logger.Debugf("using icon specified in manifest as %s", path)
+	return fmt.Sprintf("data:image/svg+xml;base64,%s", base64.StdEncoding.EncodeToString(iconData)), nil
+}
+
+func getIconDataFromPath(ctx context.Context, path string) (string, error) {
+	var icon []byte
+	if strings.HasPrefix(path, "http") {
```
lieut-data: Does this execution path matter anymore, since we have the "missing" icons always checked in?

cmd/generator/main.go
hanzei: :+1: for that suggestion. With a `database` flag in place, would there still be the possibility to output to std out? Is this needed?

cmd/generator/add.go
```diff
@@ -0,0 +1,92 @@
...
+		command.SilenceUsage = true
+
+		plugins, err := InitCommand(command)
+		if err != nil {
+			return err
+		}
+
+		repo := args[0]
+		tag := args[1]
+		if !regexp.MustCompile("v[0-9]+.[0-9]+.[0-9]+$").MatchString(tag) {
```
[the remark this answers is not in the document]
hanzei: Yes, we can. The only downside is, that `github.com/blang/semver` discards the `v` prefix and hence we can't use the returned structure in the following code.

I added it just as a checker.

cmd/generator/add.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+func init() {
+	generatorCmd.AddCommand(addCmd)
+}
+
+var addCmd = &cobra.Command{
+	Use:     "add [repo] [tag]",
+	Short:   "Add a plugin release to the plugins.json database",
+	Example: `  generator add matterpoll v1.5.1`,
```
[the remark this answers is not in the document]
hanzei: Good suggestion :+1: I added this as part of the longer help text.

cmd/generator/add.go
hanzei: Good catch, removed.

cmd/generator/main.go
hanzei: It's because it would shadow the error shadowing in L56 otherwise. Ugly, I know :confused:

cmd/generator/main.go
hanzei: See https://github.com/mattermost/mattermost-marketplace/pull/35#discussion_r368470513

cmd/generator/main.go
hanzei: At the end it doesn't make a difference as the code will return anyway next line. But I see you point and changed it.

cmd/generator/main.go
hanzei: Why is this debug message needed anyway? It should be the default case going forward.

I've removed it for now. Let me know, if you think it's valuable to keep it.

cmd/generator/main.go
hanzei: I was only moving this code around. I'm totally fine to remove it. Going forward I would also like to move more icons to the respective plugin repos. (See also https://community-release.mattermost.com/core/pl/4ekprbzjctrn9qpj8y53eifq5h)

hanzei: Would you be fine also removing the code to support non-svg icon, given that this is not supported by the server and the Marketplace only uses svg icons?

lieut-data: Yes, let's remove the non-SVG support. We'll probably have to leave the icons checked in to keep the older plugins iconified, but even this could go eventually.

cmd/generator/add.go
```diff
@@ -20,8 +20,11 @@ func init() {
 }
 
 var addCmd = &cobra.Command{
-	Use:     "add [repo] [tag]",
-	Short:   "Add a plugin release to the plugins.json database",
+	Use:   "add [repo] [tag]",
+	Short: "Add a plugin release to the plugins.json database",
+	Long: "The generator commands allows adding a specific plugin release to the database by using this command.\n\n" +
+		"The release has to be build first using the cut plugin command, which also uploads it to https://plugins-store.test.mattermost.com/release/. " +
```
lieut-data:
```suggestion
		"The release has to be built first using the /cutPlugin command, which also uploads it to https://plugins-store.test.mattermost.com/release/. " +
```

cmd/generator/add.go
```diff
@@ -20,8 +20,11 @@ func init() {
...
+		"This localtion is used to fetch the plugin release.",
```
lieut-data:
```suggestion
		"This location is used to fetch the plugin release.",
```

lieut-data:
@prapti, I'm inclined to think this PR doesn't require QA review, but is an internal tool that we'll be using to re-generate `plugins.json`, and we should be testing /that/ in the context of the marketplace instead. Let me know if you have any concerns :)

cmd/generator/main.go
hanzei: Filled https://mattermost.atlassian.net/browse/MM-21893

hanzei:
@iomodo Would you please review this PR as I have a couple of ticket next sprint that depend on this change?

cmd/generator/add.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		repo := args[0]
+		tag := args[1]
+
+		if _, err = semver.ParseTolerant(tag); err != nil {
+			return errors.Wrapf(err, "%v is an invalid tag. Something like v2.3.4 is expected", tag)
+		}
+
+		bundleURL := "https://plugins-store.test.mattermost.com/release/" + repo + "-" + tag + ".tar.gz"
```
iomodo: `test` in the URL. Not related to this PR but probably we want to remove that.

cmd/generator/main.go
```diff
@@ -120,28 +105,21 @@ var generatorCmd = &cobra.Command{
 		for _, repositoryName := range repositoryNames {
 			logger.Debugf("querying repository %s", repositoryName)
 
-			releasePlugins, err := getReleasePlugins(ctx, client, repositoryName, includePreRelease, existingPlugins)
+			var releasePlugins []*model.Plugin
+			releasePlugins, err = getReleasePlugins(ctx, client, repositoryName, includePreRelease, existingPlugins)
```
iomodo: nit: we can inline this if

cmd/generator/main.go
```diff
@@ -120,28 +105,21 @@ var generatorCmd = &cobra.Command{
...
 
 			for _, plugin := range releasePlugins {
 				if len(plugin.IconData) == 0 {
 					if iconPath, ok := iconPaths[repositoryName]; ok {
-						icon, err := getIcon(ctx, iconPath)
+						var iconData string
+						iconData, err = getIconDataFromPath(iconPath)
 						if err != nil {
 							return errors.Wrapf(err, "failed to fetch icon for repository %s", repositoryName)
 						}
```
iomodo: nit: inline if

cmd/generator/main.go
```diff
@@ -151,8 +129,7 @@ var generatorCmd = &cobra.Command{
 			}
 		}
 
-		encoder := json.NewEncoder(os.Stdout)
-		err := encoder.Encode(plugins)
+		err = json.NewEncoder(os.Stdout).Encode(plugins)
 		if err != nil {
 			return errors.Wrap(err, "failed to encode plugins result")
 		}
```
iomodo: nit: inline

cmd/generator/main.go
hanzei: Unfortunately we can't, as `err` would shadow then.

hanzei:
Merging this now as I have ticket that build on top of these changes.

@prapti Hope you are fine with this. if you want to post merge test this, let me know.

cmd/generator/main.go
hanzei: At the end it's a style question and I like the style this way :wink:

prapti:
@hanzei thank you and yes, would love to test it post merge. Adding it in my queue for testing.

hanzei:
@prapti Let me know when this is on the top of your queue. I can't provide "generic" steps for this. Please just DM me.

mickmister:
@hanzei In this PR we dropped support for SVG icons in the marketplace. Do you know why this was done?

hanzei:
@mickmister It it ready do that? According to the [manifest docs](https://developers.mattermost.com/integrate/plugins/manifest-reference/#icon_path) plugin icons are always `svg` icons. The commit message in [`bddba8d` (#35)](https://github.com/mattermost/mattermost-marketplace/pull/35/commits/bddba8de5c761e0bc94c605b8cc96536e202946c) also suggests that only `svg` icons are supported.

mickmister:
@hanzei Thanks for linking the docs that explain the requirement. I asked because I want to understand _why_ they aren't supported. @matthewbirtch and I have been discussing which icons to use and came across this requirement, and we wanted to know the background on the decision.

hanzei:
@mickmister Maybe there is a misunderstanding here. _Only_  `svg` icons are supported.

mickmister:
@hanzei I had a typo in my last comment. I meant to ask why the other image file types are not supported, and why only svg is supported.

PR #37, opened 2020-01-21
[MM-21749] [MM-22214] Add "Beta" and "Community" label

hanzei:
#### Summary
This PR allows adding a "Beta" and a "Community" label to plugin.

It currently looks like this: (JIRA is just an example)
Screenshot from 2020-02-01 09-29-10
Screenshot from 2020-02-01 09-29-14
Screenshot from 2020-02-01 09-29-18

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21749
https://mattermost.atlassian.net/browse/MM-22214


hanzei:
@asaadmahmood This is the label we talked about.

asaadmahmood:
@hanzei I think we should just have a dimmer label, like the bot label for the beta tag. The blue labels are way too prominent.

hanzei:
@asaadmahmood :+1: for a dimmer label. Do you have a suggestion?

asaadmahmood:
We can just use what we have.
So just use the Badge component and if we can't use that, then just give it these clases:
Screenshot 2020-01-27 at 2 34 46 PM

However, since we need a fixed background and color, we can give it a background of:
```
background: rgba(61, 60, 64, 0.15)
color: #444;
```

Screenshot 2020-01-27 at 2 33 54 PM

hanzei:
Screenshot from 2020-01-27 10-41-32
Screenshot from 2020-01-27 10-41-35

@asaadmahmood It looks very similar to the bot label with this setting. Would you be fine it I just re use the style of the bot label?

asaadmahmood:
Yes, we should resuse the style of the bot label, but I gave the background and color because if we use the bot label directly without giving it an inline style, the theme may overwrite the background and the color, so try it with a theme and see if the bg and color are consistent.

Also, the local tag looks quite big compared to the bot label. So we can make them of similar size.

hanzei:
I've applied the same background as the local label. This is how it looks like with a dark theme:
Screenshot from 2020-01-27 10-50-36
Screenshot from 2020-01-27 10-50-31

The background and color don't get overwritten by the theme.

asaadmahmood:
@hanzei Can I have a full page screenshot?

hanzei:
Screenshot from 2020-01-27 11-01-41

asaadmahmood:
This looks weird.
I'll get a pull and look at it locally to fix the dark theme.

cmd/generator/add.go
```diff
@@ -15,6 +15,10 @@ import (
 
 func init() {
 	generatorCmd.AddCommand(addCmd)
+
+	addCmd.Flags().Bool("beta", false, "Mark release as Beta")
+	addCmd.Flags().Bool("official", false, "Mark this plugin is maintanied by Mattermost")
```
lieut-data:
```suggestion
	addCmd.Flags().Bool("official", false, "Mark this plugin as maintained by Mattermost")
```

cmd/generator/add.go
```diff
@@ -15,6 +15,10 @@ import (
...
+	addCmd.Flags().Bool("community", false, "Mark this plugin is maintanied by the community")
```
lieut-data:
```suggestion
	addCmd.Flags().Bool("community", false, "Mark this plugin as maintained by the community")
```

cmd/generator/add.go
```diff
@@ -81,12 +103,21 @@ var addCmd = &cobra.Command{
 			return errors.Wrap(err, "failed to download plugin signature")
 		}
 
+		labels := []model.Label{}
```
lieut-data: nit:
```suggestion
		var labels []model.Label
```

internal/model/label.go
```diff
@@ -7,3 +7,20 @@ type Label struct {
...
 }
+
+var AllLabels = []Label{
+	CommunityLabel,
+	BetaLabel,
+}
+
+var CommunityLabel Label = Label{
+	Name:        "Community",
+	Description: "This plugin is maintained by the Open Source Community.",
```
hanzei: @aaronrothschild Could you please confirm that this wording is fine

cmd/generator/add.go
```diff
@@ -81,12 +103,21 @@ var addCmd = &cobra.Command{
...
+		if beta {
+			labels = append(labels, model.BetaLabel)
+		}
+
+		if official {
```
lieut-data: Should this be `community`?

internal/model/label.go
```diff
@@ -7,3 +7,20 @@ type Label struct {
...
+	URL:         "https://mattermost.com/pl/default-community-plugins",
+}
+
+var BetaLabel Label = Label{
+	Name:        "Beta",
+	Description: "This plugin is currently in Beta and it's not recommended in production use.",
```
hanzei: @aaronrothschild Same here

internal/model/label.go
```diff
@@ -7,3 +7,20 @@ type Label struct {
...
+
+var CommunityLabel Label = Label{
+	Name:        "Community",
+	Description: "This plugin is maintained by the Open Source Community",
+	URL:         "https://mattermost.com/pl/default-community-plugins",
+}
+
+var BetaLabel Label = Label{
+	Name:        "Beta",
+	Description: "This plugin is currently in Beta and it's not recommended in production use.",
```
lieut-data:
```suggestion
	Description: "This plugin is currently in beta and is not recommended for use in production.",
```

cmd/generator/add.go
hanzei: I requested the review to soon :confused:

internal/model/label.go
hanzei: I've changed it to:
`This plugin is currently in Beta and is not recommended for use in production.`

`Beta` is written with a capital `B` in the documentation (e.g. in https://docs.mattermost.com/process/definitions.html) Hence, I would also use a capital `B` here.

cmd/generator/add.go
hanzei: I think there is a noticeable nuisance here: `var labels []model.Label` will render to `null` in json, while `labels := []model.Label{}` renders to `[]` in json. I'm slightly favouring the latter as `null` might be confusing to read. 1/5

lieut-data: Ah, do we not `omitempty` for this case?

hanzei: No, we don't (https://github.com/mattermost/mattermost-marketplace/blob/master/internal/model/plugin.go#L17). We also don't `omitempty` for any other field. Hence, I did not add it for consistency.

hanzei:
@jfrerich Would you mind taking a look at this PR given your involvement into plugin reviews?

internal/model/label.go
aaronrothschild: Yes, this wording is good.

internal/model/label.go
aaronrothschild: @hanzei Both are good to go.

PR #38, opened 2020-01-21
Replace existing flag for Marketplace generator with database flag

hanzei:
#### Summary
Replace `--existing` flag with a `--database` flag that is required to run the generator. This allows us to drop `jq` and `sponge`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21893


cmd/generator/add.go
```diff
@@ -30,11 +28,22 @@ var addCmd = &cobra.Command{
...
 			return err
 		}
 
+		db, err := openDatabase(command)
+		if err != nil {
+			return errors.Wrap(err, "failed to open plugin database")
+		}
+		defer db.Close()
+
+		plugins, err := model.PluginsFromReader(db)
```
hanzei: One different approach I've considered was using the Store (https://github.com/mattermost/mattermost-marketplace/blob/master/internal/store/plugin.go)  instead. But that would require some changes to the Store and the interface was not as clean as I liked it to be.

lieut-data: I'm fine with deferring the store changes, but instead of passing around an `os.File` that requires rewinding, should we just have helper methods akin to `pluginsFromDatabase` and `pluginsToDatabase`?

cmd/generator/main.go
```diff
@@ -437,35 +451,33 @@ func getIconDataFromPath(path string) (string, error) {
...
-			return nil, errors.Wrapf(err, "failed to open existing database %s", existingDatabase)
-		}
-		defer file.Close()
-
-		existingPlugins, err = model.PluginsFromReader(file)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to read existing database %s", existingDatabase)
-		}
+	if existingDatabase == "" {
+		return nil, errors.New("database name most not be empty")
```
lieut-data:
```suggestion
		return nil, errors.New("database name must not be empty")
```

cmd/generator/main.go
```diff
@@ -437,35 +451,33 @@ func getIconDataFromPath(path string) (string, error) {
...
+	}
+	file, err := os.OpenFile(existingDatabase, os.O_RDWR, 0644)
```
lieut-data: nit: newline 
```suggestion

	file, err := os.OpenFile(existingDatabase, os.O_RDWR, 0644)
```

internal/model/plugin_test.go
```diff
@@ -84,3 +85,82 @@ func TestPluginsFromReader(t *testing.T) {
 		}, plugin)
 	})
 }
+
+func TestPluginsSetReader(t *testing.T) {
```
lieut-data:
```suggestion
func TestPluginsToWriter(t *testing.T) {
```

cmd/generator/add.go
hanzei: Good suggestion :+1:

lieut-data:
@prapti, I've marked this as not requiring tests since it's for internal tooling only.

PR #39, opened 2020-01-24
Add Gitlab plugin v1.1.0 to the marketplace

hanzei:
#### Summary
I've tested locally that the installation of v1.1.0 works.

@lieut-data I would like to also add this into the next "marketplace release" given that it has been three month since the last release and there are some great features and fixes in the release. What do you think?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22019


prapti:
@hanzei I have a concern with the plugin function itself, although not necessarily with this PR that uploads v1.1.0 to marketplace: The plugin can be enabled "successfully" even without providing the necessary details such as the oauth client id, the error for which is captured in the server log. This may in fact be how the plugin was designed in the older versions as well, but given the similar nature of the github plugin which does require the credentials before being enabled, I'm wondering if this should be addressed before putting it out there in the marketplace?

hanzei:
@prapti Thanks for sharing you concerns. @manlan, the original author of the plugin, did implement this behaviour intentionally in https://github.com/mattermost/mattermost-plugin-gitlab/commit/eea738dde7b4a30463e8585586883b2951a3c9fa#diff-7e42e9629fece6c3b56011e7de76423b. The idea being that an admin should not have to re start a plugin to have it pick up the correct config.

I see that there might be confusion for admins why the plugin doesn't work because the error ist just shown in the logs. To me the solution is to notify admins when the plugin is not correctly configured.

Given that this is not a regression and "works as designed", I'm inclined to publish the release as it is.

Let me know what you think.

saturninoabril:
(Forgive the disturbance, just checking something on label)

PR #42, opened 2020-01-31
Drop "official" label from Marketplace

hanzei:
#### Summary
I also removed the the label from the code. It's easy to dig it out again if we need it.

Screenshot from 2020-01-31 14-45-06

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22122


internal/api/label.go
```diff
@@ -20,7 +20,7 @@ func initLabels(apiRouter *mux.Router, context *Context) {
 
 // handleGetPlugins responds to GET /api/v1/labels, returning a list of all defined labels.
 func handleGetLabels(c *Context, w http.ResponseWriter, r *http.Request) {
-	response := []model.Label{model.OfficialLabel}
+	response := []model.Label{}
```
hanzei: This will be used later when the community label is added.

internal/api/labels_test.go
```diff
@@ -32,14 +32,8 @@ func TestGetLabels(t *testing.T) {
...
-	require.Len(t, respose, 1)
-
-	expectedResponse := []model.Label{{
-		Name:        "Official",
-		Description: "This plugin is maintained by Mattermost",
-		URL:         "https://mattermost.com/pl/default-community-plugins",
-		Color:       "#166de0",
-	}}
+
+	expectedResponse := []model.Label{}
```
hanzei: Same here

PR #43, opened 2020-01-31
Automatic deployment

lieut-data:
#### Summary
Update the CircleCI pipeline to automatically deploy `master` to staging and a future `production` branch to production. I've only configured AWS credentials for staging right now for testing, and thereafter we should be able to trigger a production deploy by creating the branch and merging `master` into it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-18573


Makefile
```diff
@@ -6,6 +6,7 @@ BUILD_HASH_SHORT = $(shell git rev-parse --short HEAD)
 LDFLAGS += -X "github.com/mattermost/mattermost-marketplace/internal/api.buildTag=$(BUILD_TAG)"
 LDFLAGS += -X "github.com/mattermost/mattermost-marketplace/internal/api.buildHash=$(BUILD_HASH)"
 LDFLAGS += -X "github.com/mattermost/mattermost-marketplace/internal/api.buildHashShort=$(BUILD_HASH_SHORT)"
+SLS_STAGE ?= "dev"
```
lieut-data: This change isn't strictly necessary for this PR, since I invoke `sls` manually from the CirlceCI config (too much nonsense about preserving environment variables, sudo, CircleCI, etc.), but it's useful for deploying by hand.

.editorconfig
```diff
@@ -22,6 +22,6 @@ indent_style = space
 indent_size = 4
 trim_trailing_whitespace = false
 
-[.golangci.yml]
+[*.yml]
```
lieut-data: Driving me nuts editing `.circleci/config.yml`.

.circleci/config.yml
```diff
@@ -70,9 +82,46 @@ jobs:
       - run: make build
       - check-uncommitted
       - *save_cache
+      - persist_to_workspace:
+          root: *workspace_root
+          paths:
+            - .
```
hanzei: Is there a need to copy everything over? Or is `dist/` enough?

.circleci/config.yml
```diff
@@ -1,9 +1,20 @@
 version: 2.1
 
 executors:
-  default:
+  build:
     docker:
       - image: circleci/golang:1.13
+  deploy:
+    docker:
+      - image: circleci/node:8.10
```
hanzei: Are there docker images that bring `serverless` with them?

.circleci/config.yml
```diff
@@ -70,9 +82,46 @@ jobs:
...
+          name: Deploy Function
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "master" ]; then
+                export AWS_ACCESS_KEY_ID=$AWS_STAGING_ACCESS_KEY_ID
+                export AWS_SECRET_ACCESS_KEY=$AWS_STAGING_SECRET_ACCESS_KEY
+                sudo -E sls deploy function -f server --stage staging
+            elif [ "${CIRCLE_BRANCH}" == "production" ]; then
+                export AWS_ACCESS_KEY_ID=$AWS_PRODUCTION_ACCESS_KEY_ID
+                export AWS_SECRET_ACCESS_KEY=$AWS_PRODUCTION_SECRET_ACCESS_KEY
+                sudo -E sls deploy function -f server --stage production
```
hanzei: Maybe reuse `make deploy-lambda-fast` here?

lieut-data: Right now, that target also clean builds, but I'm running this in the node container without the infrastructure needed to do same, and relying on workspace artifacts instead. Otherwise, I'd definitely agree!

.circleci/config.yml
lieut-data: I tried looking for a while, but admit to not being able to find one. Definitely would prefer that!

.circleci/config.yml
lieut-data: Probably just `dist` -- any concerns with copying everything over for simplicity?

hanzei: I'm 0/5 here. A couple of small concerns are:
- Performance
- Conflicts with other files in the future
- Copying just `dist` is "the right thing" :tm:

.circleci/config.yml
hanzei: Does https://hub.docker.com/r/softinstigate/serverless work?

.circleci/config.yml
hanzei: Good catch. Missed that :+1:

.circleci/config.yml
lieut-data: Possibly? I admit I don't know how to "trust" Docker hub images. I feel like I can trust `circleci` (since I already have to), but should I do the same for `softinstigate`?

.circleci/config.yml
lieut-data: I agree with the gist of all of that -- I'm inclined to be lazy here and "get this working", and then consider tweaking later if we find the performance unacceptable or causing conflicts. There was a bunch of churn even getting the files to share across the workspaces, that I'm hesitant to mix it up again. 2/5

lieut-data:
@prapti -- no review required for this one, since it's internal tooling relative to the deployment. (And no way /to/ test without merging.)

PR #44, opened 2020-02-03
Make use of model.IsValid()

hanzei:
#### Summary
https://github.com/mattermost/mattermost-server/pull/13609 added a `IsValid()` method to `model`. This PR makes sure every manifest in the marketplace is deemed valid my this method.

#### Ticket Link
Follow up on https://github.com/mattermost/mattermost-server/pull/13609


internal/store/store.go
```diff
@@ -35,12 +34,15 @@ func New(reader io.Reader, logger logrus.FieldLogger) (*Store, error) {
 
 func validatePlugins(plugins []*model.Plugin) error {
 	for _, plugin := range plugins {
-		if plugin.Manifest.Id == "" {
-			return errors.Errorf("plugin manifest Id is empty %+v", plugin)
+		err := plugin.Manifest.IsValid()
```
hanzei: Maybe the manifest should be even checked before the plugin gets added e.g. in `cmd/generator/main`. Thoughts?

ali-farooq0: Agree, I think we should do it in both places. The reason for being both is that when parsing in the plugins.json file we can't guarantee that the data will be valid. And we should do it when we generate because we don't want to write invalid manifest to plugins.json

internal/store/store.go
```diff
@@ -35,12 +34,15 @@ func New(reader io.Reader, logger logrus.FieldLogger) (*Store, error) {
...
+		if err != nil {
+			return errors.Wrapf(err, "invalid manifest for plugin %s", plugin.Manifest.Id)
 		}
-		if _, err := semver.Parse(plugin.Manifest.Version); err != nil {
-			return errors.Wrapf(err, "failed to parse manifest version for manifest.Id %s", plugin.Manifest.Id)
+
+		if plugin.Manifest.Version == "" {
```
ali-farooq0: ahhh i think this will break things, we have manifest that does not contain `version`. What we need to check here is if a version is present, is it valid, if there is no version that that is valid too.

hanzei: The previous code did the same: confirm a version is present and error otherwise. There is even a test for this: https://github.com/mattermost/mattermost-marketplace/pull/44/files#diff-b30f396f3719b724710bcd34e4bcb1aaR36

The code is just keeping this behaviour. Are you suggesting to change it?

ali-farooq0: oh whoops, i was thinking `MinServerVersion`. You're right we should always have a version value.

Hmm, is `manifest.version` optional? i thought that field is mandatory? if it is mandatory shouldn't `Manifest.IsValid()` check if the version is there? rather than doing this check here?

hanzei: @ali-farooq0 Well, there is no technical requirement for `manifest.version` right now. If a plugin omits it, the UI just looks a bit weird. 
Screenshot from 2020-02-05 10-52-23

Why should it be (technical) mandatory?

hanzei:
I'm going to merge this PR given that it just preserves the existing behaviour. @ali-farooq0 I'm happy to continue the discussion about what `model.IsValid()` should check or not.

internal/store/store.go
ali-farooq0: @hanzei 
that's our only frame of reference to see if a plugin is newer or not, if an author keeps releasing with no version we'll keep overwriting existing plugins in marketplace, no? maybe that could be a requirement for marketplace.

PR #46, opened 2020-02-04
document automatic deployment

lieut-data:
#### Summary
Document the new automatic deployment process for the marketplace.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-18573
https://mattermost.atlassian.net/browse/MM-18574


README.md
```diff
@@ -51,14 +51,24 @@ make plugins.json
 
 In addition to running as a standalone server, the Marketplace is also designed to run as a Lambda function, compiling the `plugins.json` database into the binary for immediate access without further configuration.
 
+### Automatic Deployment
+
+Changes merged to `master` are automatically deployed to https://api.staging.integrations.mattermost.com.
+
+Changes merged to `production` are automatically deployed to https://api.integrations.mattermost.com.
+
+To update production, simply create, review and merge the appropriate pull request. Often this will just be a merge from `master` to `production`, but if there are pending changes not yet ready for a production release there, it may be appropriate to just update the plugin database with a direct change to production that is later merged back to `master`.
```
hanzei: I would like to avoid having to merge `master` into `production` all to often. I was envisioning ~one merge per month.

Why should updates to `plugins.json` not always to to `production`?

hanzei: In my original proposal the wording was:
> Adding new releases to plugins or adding a new plugin should be submitted to `production`. **Any** other changes should go into `master`.

What I like about this is that it very clear, where every change should go.

ali-farooq0: feel like once a month is slow, unless ofcourse we're pushing `plugins.json` more often directly to production then 👌

lieut-data: Tweaked the proposal here -- thoughts?

lieut-data:
@hanzei & @ali-farooq0, thoughts on these changes?

README.md
```diff
@@ -51,14 +51,24 @@ make plugins.json
...
+
+When adding or updating the plugins database (or corresponding tooling), submit the changes directly to `production`, and then merge `production` immediately back to `master` to reduce unnecessary merge conflicts. All other changes should be committed directly to `master`. Changes pending release from `master` should be merged to `production` after qualification and in coordination with any supporting Mattermost server release.
+
+### Manual Deployment
+
 Simply run the following:
 
 ```
-$ make deploy-lambda
+$ SLS_STAGE=staging make deploy-lambda
```
ali-farooq0: for manual production deployment that may need to happen for any reason, will be done through AWS?

lieut-data: Apologies, I'm not sure I fully understand the question -- if the auto deploy fails, this can be done via serverless with the appropriate AWS credentials, yes.

ali-farooq0: Sorry, yes was wondering if auto-deploy fails, how would we deploy to production in that case. Cool looks like we should be able to do that using AWS. 👍

README.md
```diff
@@ -51,14 +51,24 @@ make plugins.json
 
 In addition to running as a standalone server, the Marketplace is also designed to run as a Lambda function, compiling the `plugins.json` database into the binary for immediate access without further configuration.
 
+### Automatic Deployment
+
+Changes merged to `master` are automatically deployed to https://api.staging.integrations.mattermost.com.
+
+Changes merged to `production` are automatically deployed to https://api.integrations.mattermost.com.
+
+When adding or updating the plugins database (or corresponding tooling), submit the changes directly to `production`, and then merge `production` immediately back to `master` to reduce unnecessary merge conflicts. All other changes should be committed directly to `master`. Changes pending release from `master` should be merged to `production` after qualification and in coordination with any supporting Mattermost server release.
```
hanzei: Is there a difference between merging `production` to `master` and cherry-picking the lastest commit from `production` into `master`?

lieut-data: The only difference in my mind is that the former catches anything that might have been missed, forcing an earlier reconciliation. If we wanted to go down the cherry-pick route, I'd suggest we adopt the mattermost-server semantics, commit everything to `master`, and cherry-pick to production instead. Thoughts?

(I'll merge this regardless, and we can iterate as needed.)

PR #49, opened 2020-03-19
Merge master into production

hanzei:
#### Summary
This is needed so `production` has benefit from https://github.com/mattermost/mattermost-marketplace/pull/47 and #48, which are needed to fix the build pipeline.


hanzei:
@prapti Heads up that this is the same version/commit that ` api.staging.integrations.mattermost.com` is using. Hence you can use this instance for testing.

prapti:
> @prapti Heads up that this is the same version/commit that ` api.staging.integrations.mattermost.com` is using. Hence you can use this instance for testing.

@hanzei is a smoke test of marketplace needed for this PR or what specific testing was needed?

hanzei:
@prapti This PR merges the latest changes on `master` into `production`. Hence, every change in this PR has been reviewed individuality. A smoke test is fine.

PR #52, opened 2020-04-01
Add icon for Zoom plugin

hanzei:
#### Summary
Screenshot from 2020-04-01 07-26-54

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-zoom/issues/94


cmd/generator/main.go
```diff
@@ -86,18 +86,18 @@ var generatorCmd = &cobra.Command{
...
 			"mattermost-plugin-aws-SNS",
+			"mattermost-plugin-custom-attributes",
+			"mattermost-plugin-github",
 			"mattermost-plugin-gitlab",
+			"mattermost-plugin-jira",
+			"mattermost-plugin-jenkins",
 			"mattermost-plugin-nps",
 			"mattermost-plugin-webex",
+			"mattermost-plugin-welcomebot",
+			"mattermost-plugin-zoom",
```
hanzei, line 100: I did took the opportunity to order the repositories alphabetically.

hanzei:
Head up that this PR goes directly into production.

cmd/generator/main.go
```diff
@@ -107,6 +107,7 @@ var generatorCmd = &cobra.Command{
 			"mattermost-plugin-jenkins": "data/icons/jenkins.svg",
 			"mattermost-plugin-jira":    "data/icons/jira.svg",
 			"mattermost-plugin-webex":   "data/icons/webex.svg",
+			"mattermost-plugin-zoom":    "data/icons/zoom.svg",
```
lieut-data, line 110: Hosting the images locally was just meant as a stop-gap until the plugins defined the icons themselves, and posted a release containing same. Could we just wait for that, and slowly reduce this list to zero instead?

hanzei: Given that there are no "old" release of the zoom plugin that need the icon, I agree that it's not worth continuing to use this workaround. I will close the PR.

PR #54, opened 2020-04-03
Use GITHUB_TOKEN env variable instead of having the user define it

hanzei:
#### Summary
Otherwise `make plugins.json` will fail if the user hasn't set `GITHUB_TOKEN`.

I also updated the documentation.


hanzei:
@jfrerich This fixes the issue you did encounter

hanzei:
@prapti I think the best way to debug this is to make run `make plugins.json` works fine with and without `GITHUB_TOKEN` set.

hanzei:
@prapti Just want to touch bases: Do you have any questions about how to test this PR?

prapti:
Hi @hanzei sorry my reviews are taking longer this time around if they're outside 5.22 since we're wrapping up some final testing for the release. I'll get on the reviews here soon. Thank you for the patience, and apologies for the lateness!

hanzei:
Thanks for the review @prapti :+1:

PR #60, opened 2020-04-23
Proxy upstream support

lieut-data:
#### Summary
Support proxying to an upstream marketplace via `--upstream` or the `BUILD_UPSTREAM_URL` environment variable that is baked into a binary.

This is useful for customers that want to supply their own plugins, but simultaneously stay up-to-date to an upstream marketplace like `api.integrations.mattermost.com`. This is immediately useful for launching a "beta listing" of the marketplace in which to list upcoming efforts, such as the incident response plugin, but without exposing to all users of the marketplace.

#### Ticket Link
N/A


internal/store/proxy_test.go
```diff
@@ -0,0 +1,78 @@
...
+		require.NoError(t, err)
+
+		plugins, err := proxyStore.GetPlugins(&model.PluginFilter{
+			PerPage: model.AllPerPage,
+		})
+		require.NoError(t, err)
+		require.Empty(t, plugins)
+	})
+
+	t.Run("empty stream", func(t *testing.T) {
```
ali-farooq0, line 34: `"empty stream with error"` to differentiate from the last test

lieut-data:
```suggestion
	t.Run("empty stream with error", func(t *testing.T) {
```

Makefile
```diff
@@ -52,7 +53,7 @@ run: run-server
 
 ## Run the Plugin Marketplace
 .PHONY: run-server
-run-server:
+run-server: generate
```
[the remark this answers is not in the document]
hanzei: `github.com/mattermost/mattermost-marketplace/data/statik` is only imported in 
https://github.com/mattermost/mattermost-marketplace/blob/856bd9d7fd1cb9a8edc69a52d09bd1a9802a47ec/cmd/lambda/main.go#L10. A "normal" server doesn't use it, right? I'm don't quite understand why re-generating is needed.

cmd/lambda/main.go
```diff
@@ -45,14 +50,26 @@ func newStatikStore(statikPath string, logger logrus.FieldLogger) (*store.Store,
...
+		upstreamStore, err := store.NewProxy(upstreamURL, logger)
+		if err != nil {
+			return errors.Wrap(err, "failed to initialize upstream store")
+		}
+
+		stores = append(stores, upstreamStore)
+	}
 
 	router := mux.NewRouter()
 	api.Register(router, &api.Context{
```
hanzei, line 71: It seams counter intuitive to to generate a `store.NewMerged` using only one `store`. 

What about declaring a `store.Store` first and then make it an `MergedStore` if needed:
```suggestion
	var newStore store.Store
	newStore, err := newStatikStore("/plugins.json", logger)
	if err != nil {
		return err
	}

	if upstreamURL != "" {
		upstreamStore, err := store.NewProxy(upstreamURL, logger)
		if err != nil {
			return errors.Wrap(err, "failed to initialize upstream store")
		}

		newStore = store.NewMerged(logger, newStore, upstreamStore)
	}

	router := mux.NewRouter()
	api.Register(router, &api.Context{
		Store:  newStore,
		Logger: logger,
	})
```

Makefile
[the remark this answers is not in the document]
lieut-data: You're right -- I'm not sure what problem this was solving. Reverting for now.

DHaussermann:
@lieut-data can you please provide some guidance for testing this change? It's not quite clear to me how to test this. 
If start my server with`go run ./cmd/marketplace server --upstream https://api.integrations.mattermost.com` and then point the Mattermost server to my local instance, should it also show me additional plgins coming from `https://api.integrations.mattermost.com`?

lieut-data:
@DHaussermann, yes, that's basically the idea. You'll have to edit your local `plugins.json` with something to "overlay" the upstream. Note that this is live on https://api.private-beta.integrations.mattermost.com/ as well.

DHaussermann:
@hanzei if you have time, is this something you would be able to peer test using the peer testing process documented here?https://mattermost.atlassian.net/wiki/spaces/INT/pages/619937892/Peer+Testing+Process

hanzei:
Sure @DHaussermann :+1:

hanzei:
@DHaussermann :point_up:

DHaussermann:
Thanks for testing @hanzei! Much appreciated.

Issue #61, opened 2020-04-23
Drop iconPaths workaround


hanzei:
Hosting the images locally was just meant as a stop-gap until the plugins defined the icons themselves, and posted a release containing same. Could we just wait for that, and slowly reduce this list to zero instead?

_Originally posted by @lieut-data in https://github.com/mattermost/mattermost-marketplace/pull/52_

PR #62, opened 2020-04-23
Drop iconPaths workaround

hanzei:
#### Summary
Given that all plugin listed in `iconPaths` have there icon now set via `icon_path` in there manifest it's not longer needed to have the icon in this repository. All future releases of the plugins will have an icon.

Please note that this changes doesn't effect existing releases. There is no `diff` for `plugins.json`.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-marketplace/issues/61


ali-farooq0:
What about older MM servers though? 🤔

ali-farooq0:
Ahh right, this is just the generation. `plugin.json` still has the icons. Forgot about that. Thanks for the explanation @lieut-data 👍

PR #68, opened 2020-05-25
Ensure manually added plugins stay in the database

hanzei:
#### Summary
Prior to this PR the generator would drop all manually added plugins. These changes ensure the generator only adds new plugins and never removes old ones.
 This also means that removing a GH release does not longer drop the plugin from the Marketplace. I tried preserving the old behaviour, but I didn't found an elegant way. 1/5 that it's not worth fixing this, as that case should rarely happend.

#### QA Steps
1. Run `go run ./cmd/generator/ add mattermost-plugin-todo v0.2.0 --beta --official`
2. Ensure via `git diff` that Todo plugin was added
3. Run `make plugins.json`
4. Ensure via `git diff` that Todo plugin still exists in `plugins.json`

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-marketplace/issues/67


cmd/generator/main.go
```diff
@@ -112,6 +112,23 @@ var generatorCmd = &cobra.Command{
...
+					found = true
+				}
+			}
+
+			if !found {
+				manuallyAdded = append(manuallyAdded, ep)
+			}
+		}
+
+		plugins = append(plugins, manuallyAdded...)
```
hanzei, line 130: I would really like to sort all plugins first by `id` and then by `version`. That way manually added plugins would not always be added to the end of the list. But that change would generate a huge diff. Thoughts?

larkox: Is this the order of the user facing list? I think that ordering by ID might not be optimal, since IDs might not end up alphabetically ordered as expected (i.e. I would like to have plugin "D", before plugin "E", but if the ids are "d" and "com.a.e", "E" will be listed first). I would try a more complex idea:
1. Group plugins by ID
2. Order the plugin inside each group by version.
3. Take the plugin Name of the highest version plugin and make that a "representative or the group".
4. Order the groups by the representative Name alphabetically.

That way, if I change the plugin name from `My awesome plugin` to `Awesome plugin` both will be together, and in the order of the one of the highest version. It will slightly break the alphabetic order, but it feels better for me.

lieut-data: No, this output does should not affect the user-facing list: it was intended just for hygiene in the generated file.

I'm fine with a one-time diff this time around, @hanzei, if it simplifies things going forward.

hanzei: @lieut-data @larkox I did add the sorting method. PTAL.

larkox:
Sure, will do.

larkox:
@DHaussermann 
Scope
(Your understanding of what functional impact the PR will have on the product)
---------
During the creation of plugins.json, manually added plugins should remain in the list.


Testing notes
(Testing you've done that you feel validate the change and cover any needed regresssion) 
----------------
- Followed the QA steps provided. Everything worked fine.
- Repeated the step 3. The status of plugins.json is the same (does not create duplicates)

Follow-up concerns
------------------- 
No concerns on my side.

PR #73, opened 2020-06-02
Add v1.2.0 of mattermost-plugin-confluence to the Marketplace

jfrerich:
#### Summary
Add confluence v1.2.0 to marketplace

Note that confluence is a new marketplace plugins and needed to be added to the generate script.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-confluence/issues/51


cmd/generator/main.go
```diff
@@ -98,15 +98,17 @@ var generatorCmd = &cobra.Command{
 			"mattermost-plugin-gitlab",
 			"mattermost-plugin-nps",
 			"mattermost-plugin-webex",
+			"mattermost-plugin-confluence",
```
lieut-data, line 101: This part of the generator is legacy -- just meant to support plugins that don't have exhaustively defined manifests including an icon. I think we can revert as such, assuming the `confluence.svg` is part of the confluence plugin repository. @hanzei, can we maybe just rip out all this code now?

jfrerich: Thanks Jesse.  I noticed this section when looking at the array plugin names defined above.  If you and @hanzei agree this is no longer needed, I can remove this icon statement and the icon from the `data/icons` directory

hanzei: I agree that this is a legacy part of the code. Let's not add new plugins to it. I'm 0/5 if it's worth removing right now.

jfrerich:
@hanzei, @lieut-data I reverted adding the icon SVG file and the entry from the icon struct.  

Regarding the conflicting `data/statik/statik.go` issue.  What the correct resolution there?  The file is going to be overwritten, correct?

lieut-data:
@jfrerich, whatever the generator produces should override `statik.go`, so feel free to resolve in favour of your changes.

jfrerich:
@lieut-data, @DHaussermann, @levb we are good to go again, with a fresh commit

Issue #77, opened 2020-06-16
Only show EE-only plugins if enterprise_plugins is set


hanzei:
Follow up from https://github.com/mattermost/mattermost-server/pull/14586

#### QA Test steps
**Setup**
1. Set Marketplace URL to `http://localhost:8085`
2. Enable Remote Marketplace
3. Add an EE-only plugin by running `go run ./cmd/generator/ add mattermost-plugin-mscalendar v1.0.0 --official --enterprise` in Marketplace repo
4. Start local Marketplace using `make run`

**1. Test**
1. Run Mattermost Server Version `v5.23.0` with E20 license
2. Confirm that Markplace model **does not** show Microsoft Calendar Plugin. The plugin can be successfully installed.

**2. Test**
1. Run Mattermost Server Version `v5.24.0` without a license
2. Confirm that Markplace model **does** show Microsoft Calendar Plugin. The plugin can be successfully installed.

**3. Test**
1. Run Mattermost Server Version `v5.24.0` with a E10 license
2. Confirm that Markplace model **does** show Microsoft Calendar Plugin. The plugin can be successfully installed.

**4. Test**
1. Run Mattermost Server Version `v5.24.0` with a E20 license
2. Confirm that Markplace model **does** show Microsoft Calendar Plugin. The plugin can be successfully installed.

**5. Test**
1. Run Mattermost Server from https://github.com/mattermost/mattermost-server/pull/14913 without a license
2. Confirm that Markplace model **does not** show Microsoft Calendar Plugin

**6. Test**
1. Run Mattermost Server from https://github.com/mattermost/mattermost-server/pull/14913 with a E10 license
2. Confirm that Markplace model **does not** show Microsoft Calendar Plugin

**7. Test**
1. Run Mattermost Server from https://github.com/mattermost/mattermost-server/pull/14913 with a E20 license
2. Confirm that Markplace model **does** show Microsoft Calendar Plugin. The plugin can be successfully installed.

PR #79, opened 2020-06-17
Limit enterprise plugins to E20 installations

hanzei:
#### Summary
Allow plugins to be added as "Enterprise" plugins. These will only be returned if the server sends `enterprise_plugins` as `true`. Given that that flag was added in `v5.24`, all server below `v5.24` will not be to install enterprise plugins via the Marketplace.

I added an `Enterprise` label to inform admins that a downgrade of there license will stop the plugin from working. 0/5 on the wording of the tooltip.

Screenshot from 2020-06-18 10-25-58

For the full list of test cases please see https://github.com/mattermost/mattermost-marketplace/issues/77

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-marketplace/issues/77


internal/model/label.go
```diff
@@ -24,3 +24,9 @@ var BetaLabel Label = Label{
 	Description: "This plugin is currently in Beta and is not recommended for use in production.",
 	URL:         "https://mattermost.com/pl/default-beta-plugins",
 }
+
+var EnterpriseLabel Label = Label{
+	Name:        "Enterprise",
+	Description: "This plugin only works on servers with an E20 license.",
+	URL:         "https://mattermost.com/pl/TODO",
```
hanzei, line 31: @aaronrothschild Should there be a link to some documentation page? Or is no link needed?

internal/model/label.go
```diff
@@ -24,3 +24,9 @@ var BetaLabel Label = Label{
 	Description: "This plugin is currently in Beta and is not recommended for use in production.",
 	URL:         "https://mattermost.com/pl/default-beta-plugins",
 }
+
+var EnterpriseLabel Label = Label{
+	Name:        "Enterprise",
+	Description: "This plugin only works on servers with an E20 license.",
```
hanzei, line 30: @aaronrothschild What do you think about this wording?

internal/model/plugin.go
```diff
@@ -17,7 +17,8 @@ type Plugin struct {
 	Labels          []Label                   `json:"labels,omitempty"`
 	Signature       string                    `json:"signature"` // A signature of a plugin saved in base64 encoding.
 	Manifest        *mattermostModel.Manifest `json:"manifest"`
-	UpdatedAt       time.Time                 `json:"updated_at"` // The point in time this release of the plugin was added to the Plugin Marketplace
+	Enterprise      bool                      `json:"enterprise,omitempty"` // Indicated if the plugin is an enterprise plugin
```
hanzei, line 20: Using `omitempty` ensures there is no diff in `plugins.json`. On the other hand, not omitting this field might be cleaner, as one can see at first glance, what type of plugin it is.

JtheBAB:
Should the flag not be backported to the latest ESR? I guess the targeted audience is more on ESR then on the latest version?

internal/model/label.go
aaronrothschild:
```suggestion
	URL:         "https://mattermost.com/pricing/",
```

cmd/generator/add.go
```diff
@@ -17,8 +17,9 @@ func init() {
 	generatorCmd.AddCommand(addCmd)
 
 	addCmd.Flags().Bool("beta", false, "Mark release as Beta")
-	addCmd.Flags().Bool("official", false, "Mark this plugin as maintanied by Mattermost")
-	addCmd.Flags().Bool("community", false, "Mark this plugin as maintanied by the Open Source Community")
+	addCmd.Flags().Bool("official", false, "Mark this plugin as maintained by Mattermost")
+	addCmd.Flags().Bool("community", false, "Mark this plugin as maintained by the Open Source Community")
+	addCmd.Flags().Bool("enterprise", false, "Mark this plugin as only available to enterprise installations")
```
aaronrothschild, line 22:
```suggestion
	addCmd.Flags().Bool("enterprise", false, "Mark this plugin as only available to installations with E20-only plugins license")
```

aaronrothschild: This then explains what the check is actually looking for

internal/model/label.go
```diff
@@ -24,3 +24,9 @@ var BetaLabel Label = Label{
 	Description: "This plugin is currently in Beta and is not recommended for use in production.",
 	URL:         "https://mattermost.com/pl/default-beta-plugins",
 }
+
+var EnterpriseLabel Label = Label{
+	Name:        "Enterprise",
```
aaronrothschild, line 29:
```suggestion
	Name:        "Enterprise (E20)",
```

aaronrothschild: This then indicates the SKU needed for it to operate.

internal/model/label.go
aaronrothschild, line 31:
```suggestion
	URL:         "https://mattermost.com/pricing/",
```

internal/model/label.go
aaronrothschild: I think it's good and clear

hanzei:
> Should the flag not be backported to the latest ESR? I guess the targeted audience is more on ESR then on the latest version?

Mscalender requires `v5.24.0` to be run. Hence, the changes in this PR also only target Mattermost Server `v5.24.0`.

internal/store/static.go
```diff
@@ -114,11 +114,16 @@ func (store *StaticStore) GetPlugins(pluginFilter *model.PluginFilter) ([]*model
 }
 
 // getPlugins returns all plugins compatible with the given server version, sorted by name ascending.
-func (store *StaticStore) getPlugins(serverVersion string) ([]*model.Plugin, error) {
+// Only returns enterprise plugins, if enterprisePlugins is true.
```
lieut-data, line 117: I initially read this as, "if enterprisePlugins is true, only return enterprise plugins", but of course that's not the semantics. Could we just rename the variable to `includeEnterprisePlugins` and eliminate this comment (and it's ambiguity)?

internal/api/plugins.go
```diff
@@ -33,11 +33,17 @@ func parsePluginFilter(u *url.URL) (*model.PluginFilter, error) {
 	filter := u.Query().Get("filter")
 	serverVersion := u.Query().Get("server_version")
 
+	enterprisePlugins, err := parseBool(u, "enterprise_plugins", false)
```
ali-farooq0, line 36: I wonder if we can shorten this to just `enterprise`. But then maybe it becomes ambiguous between do i include enterprise plugins or do i only return enterprise plugins. What about `include_enterprise=true|false`?

internal/model/plugin.go
ali-farooq0: Personally I like being explicit with fields and would want to get rid of `omitempty` here -- the `plugins.json` diff noise will only happen once so personally don't mind that. But 0/5 on this since it only affects `plugins.json`.

internal/api/plugins.go
hanzei: `enterprise_plugins` was defined and added in https://github.com/mattermost/mattermost-server/pull/14586, which shipped with `v5.24.0`. I'm 3/5 that `model.PluginFilter` should be a "copy" of http request. I could see value in changing the field in `GetPluginsRequest` to something like `include_enterprise`. But consistency between `GetPluginsRequest` and  `model.PluginFilter` seams more important to me. 1/5

internal/store/static.go
hanzei: Good suggestion :+1:

internal/model/plugin.go
hanzei: :+1: for being explicit

internal/api/plugins.go
ali-farooq0: Ahh didn't know that webapp was already using this field. Agree that it's better to be consistent in that regard.

DHaussermann:
Tested following all the steps provided in linked issue.
- The  plugin does correctly only display when the version is 5.24 **AND** the licencee is E20 🎉 

@hanzei  or @lieut-data  is the scope of this PR?  Just to confirm something here. When the plugin does appear and I click the button to install the link it's pointing top appears to be wrong.
MarketPlace

Just to confirm - This PR is only to validate the display behavior and it's expected that the button is not functional yet?

hanzei:
@DHaussermann That is a very valid find of a bug, which has to be fixed before this PR gets merged.

The root cause is, that I forgot in https://github.com/mattermost/mattermost-server/pull/14586 that only `App.getRemotePlugins` sends the `EnterprisePlugins` flag (https://github.com/mattermost/mattermost-server/blob/6158e91308f2d2255a88c680dfca3ac439dbf11a/app/plugin.go#L509-L529), but not `App.getRemoteMarketplacePlugin`(https://github.com/mattermost/mattermost-server/blob/6158e91308f2d2255a88c680dfca3ac439dbf11a/app/plugin.go#L491-L500). Hence, listing the plugin with an E20 license works, but not installing it, because the Marketplace assumes that a non-E20 instance makes the request.

While this bug can be fixed in Mattermost `v5.25.0`, `v5.24.0` is already cut and released. The only "fix" I can think about that we can do in the Marketplace is to check if `Filter` is set. Because only requests generated in `App.getRemoteMarketplacePlugin` set `Filter`. Therefore the Marketplace should return Enterprise plugins even if `EnterprisePlugins` is not set, if `Filter != ""`. That is a hacky workaround (which might need to be guarded by `if pluginFilter.ServerVersion == "5.24.0`, but it would fix the bug found by @DHaussermann.

@ali-farooq0 @lieut-data What do you think about this idea? Is it too hacky?

lieut-data:
@hanzei, for simplicity, I'd propose we fix this in v5.25 and only start responding to this bit when the reported server version matches same.

ali-farooq0:
I like @lieut-data's idea. Simple with low risk. But it does mean that this enterprise feature will be delayed and will work "correctly" on a later mattermost-server version, which isn't too bad.

internal/store/static.go
```diff
@@ -114,20 +114,33 @@ func (store *StaticStore) GetPlugins(pluginFilter *model.PluginFilter) ([]*model
...
 // getPlugins returns all plugins compatible with the given server version, sorted by name ascending.
-func (store *StaticStore) getPlugins(serverVersion string) ([]*model.Plugin, error) {
+func (store *StaticStore) getPlugins(serverVersion string, includeEnterprisePlugins bool) ([]*model.Plugin, error) {
 	var result []*model.Plugin
 	plugins := map[string]*model.Plugin{}
 
 	for _, storePlugin := range store.plugins {
-		if serverVersion != "" && storePlugin.Manifest.MinServerVersion != "" {
-			meetsMinServerVersion, err := storePlugin.Manifest.MeetMinServerVersion(serverVersion)
+		if serverVersion != "" {
```
hanzei, line 122: I thought about handling the case where `serverVersion == ""`. But the complexity didn't justified that value to me. The case `serverVersion == ""` should never happen with the mattermost server, right?

hanzei:
@lieut-data @ali-farooq0 I did update the PR to include the workaround for https://mattermost.atlassian.net/browse/MM-26507. Please take another look. See also the server PR https://github.com/mattermost/mattermost-server/pull/14913

internal/store/static.go
```diff
@@ -78,7 +78,7 @@ func (store *StaticStore) GetPlugins(pluginFilter *model.PluginFilter) ([]*model
 		return nil, nil
 	}
 
-	plugins, err := store.getPlugins(pluginFilter.ServerVersion)
+	plugins, err := store.getPlugins(pluginFilter.ServerVersion, pluginFilter.EnterprisePlugins)
```
lieut-data, line 81: It's still weird to me that the filter parameter is `EnterprisePlugins` but then we go ahead and call it `includeEnterprisePlugins` in the receiving function. Worth converging on the same terminology across the board?

internal/store/static.go
```diff
@@ -114,20 +114,33 @@ func (store *StaticStore) GetPlugins(pluginFilter *model.PluginFilter) ([]*model
...
+		if serverVersion != "" {
+			sv, err := semver.Parse(serverVersion)
 			if err != nil {
-				return nil, errors.Wrapf(err, "failed to check minServerVersion for manifest.Id %s", storePlugin.Manifest.Id)
+				return nil, errors.Wrapf(err, "failed to parse serverVersion %s", serverVersion)
 			}
 
-			if !meetsMinServerVersion {
+			// Workaround for https://mattermost.atlassian.net/browse/MM-26507
+			if sv.GE(semver.MustParse("5.25.0")) && storePlugin.Enterprise && !includeEnterprisePlugins {
```
lieut-data, line 129: Can we extract `semver.MustParse("5.25.0")` as a "constant" named `minVersionSupportingEnterpriseFlags`?

internal/store/static.go
lieut-data: In my opinion, the marketplace filters should work for non-mattermost-server clients. I often run a curl against the marketplace to check on the status, and I'd expect the `EnterprisePlugins` flag to work.

internal/store/static.go
```diff
@@ -114,20 +114,33 @@ func (store *StaticStore) GetPlugins(pluginFilter *model.PluginFilter) ([]*model
...
-			meetsMinServerVersion, err := storePlugin.Manifest.MeetMinServerVersion(serverVersion)
+		if serverVersion != "" {
+			sv, err := semver.Parse(serverVersion)
 			if err != nil {
-				return nil, errors.Wrapf(err, "failed to check minServerVersion for manifest.Id %s", storePlugin.Manifest.Id)
+				return nil, errors.Wrapf(err, "failed to parse serverVersion %s", serverVersion)
 			}
 
-			if !meetsMinServerVersion {
+			// Workaround for https://mattermost.atlassian.net/browse/MM-26507
```
ali-farooq0, line 128: Mind putting another comment something like: 
`// Honour enterprise flag for server version >= 5.25.0 only.`
just for quicker understanding of why we need to do this.

internal/store/static.go
hanzei: Good suggestion :+1:

internal/store/static.go
hanzei: :+1: Added

internal/store/static.go
```diff
@@ -119,30 +119,35 @@ func (store *StaticStore) getPlugins(serverVersion string, includeEnterprisePlug
...
-			if sv.GE(semver.MustParse("5.25.0")) && storePlugin.Enterprise && !includeEnterprisePlugins {
+		if storePlugin.Enterprise && !includeEnterprisePlugins {
+			if serverVersion == "" {
 				continue
-			}
-
-			if storePlugin.Manifest.MinServerVersion != "" {
-				var meetsMinServerVersion bool
-				meetsMinServerVersion, err = storePlugin.Manifest.MeetMinServerVersion(serverVersion)
+			} else {
```
lieut-data, line 125: nit: Can we just eliminate the `else` and outdent, given the `continue` above?

internal/store/static.go
hanzei: Respecting the `curl` use case makes sense. I did updated the code.

internal/store/static.go
```diff
@@ -119,30 +119,35 @@ func (store *StaticStore) getPlugins(serverVersion string, includeEnterprisePlug
...
+				sv, err := semver.Parse(serverVersion)
 				if err != nil {
-					return nil, errors.Wrapf(err, "failed to check minServerVersion for manifest.Id %s", storePlugin.Manifest.Id)
+					return nil, errors.Wrapf(err, "failed to parse serverVersion %s", serverVersion)
 				}
 
-				if !meetsMinServerVersion {
+				// Honor enterprise flag for server version >= 5.25.0 only.
+				// Workaround for https://mattermost.atlassian.net/browse/MM-26507
+				minVersionSupportingEnterpriseFlags := semver.MustParse("5.25.0")
```
lieut-data, line 133: nit: can we define this at the top of the file for clarity, vs. running it each time for each plugin? (Note, it's not a performance concern, just clarity.)

internal/store/static.go
hanzei: My reasoning behind the naming was that from the perspective of the server `EnterprisePlugins` only indicates that the server has the license to run Enterprise Plugins. He decision to also deliver these plugins is done my the Marketplace and hence the name here is `includeEnterprisePlugins`. In the future, the Marketplace might react differently.

internal/store/static.go
hanzei: I'm not sure, if that would help with clarity. To me a global variable indicates that it's used in multiple methods, which is not the case here.

How would moving this into a global scope make the code more clear?

lieut-data: I don't think it's strictly necessary for a global variable to be used in multiple functions, but I do see your point. It's just /definitely/ weird to me have to keep declaring `minVersionSupportingEnterpriseFlags` on each iteration of the loop. Pulling this to the top of the file neatly decouples this, and likely fits in with other such minimum versions that will inevitably arise.

internal/store/static.go
```diff
@@ -114,11 +114,29 @@ func (store *StaticStore) GetPlugins(pluginFilter *model.PluginFilter) ([]*model
...
 
 // getPlugins returns all plugins compatible with the given server version, sorted by name ascending.
-func (store *StaticStore) getPlugins(serverVersion string) ([]*model.Plugin, error) {
+func (store *StaticStore) getPlugins(serverVersion string, includeEnterprisePlugins bool) ([]*model.Plugin, error) {
 	var result []*model.Plugin
 	plugins := map[string]*model.Plugin{}
 
 	for _, storePlugin := range store.plugins {
+		if storePlugin.Enterprise && !includeEnterprisePlugins {
+			if serverVersion == "" {
```
ali-farooq0, line 123: `serverVersion == ""` means no server version restriction, right? in which case shouldn't we include enterprise plugins? as we do for server versions < 5.25.0?

lieut-data: I think this is the case where we exclude enterprise plugins if requested by a client that doesn't specify a server version (i.e. make the `EnterprisePlugins` flag "real")

ali-farooq0: Ahh right, in this case we _should_ be honouring the flag.

hanzei: > I think this is the case where we exclude enterprise plugins if requested by a client that doesn't specify a server version (i.e. make the `EnterprisePlugins` flag "real")

Exactly

internal/store/static.go
hanzei: Good point on the potential other min versions. I did move it on top.

hanzei:
> Just two nits below. Last question: since we ignore v5.24's request to filter the enterprise plugins, I guess that means that unlicensed v5.24 will see enterprise plugins by default, only for them to disappear on upgrade to v5.25?
> 
> Would it make sense to force `EnterprisePlugins = false` when `serverVersion <= v5.24`?

Setting `EnterprisePlugins = false` when `serverVersion <= v5.24` would block existing v5.24 installations from installing MS calender. That doesn't seam desirable to me.

lieut-data:
> Setting EnterprisePlugins = false when serverVersion <= v5.24 would block existing v5.24 installations from installing MS calender. That doesn't seam desirable to me.

Got it. So long as it's explicitly understood that that plugin will appear in v5.24 and then disappear in v5.25. Thanks for confirming :)

hanzei:
> Got it. So long as it's explicitly understood that that plugin will appear in v5.24 and then disappear in v5.25. Thanks for confirming :)

Just checked with @aaronrothschild. This is the correct behaviour.

DHaussermann:
Tested and passed 
This was successfully tested together with https://github.com/mattermost/mattermost-server/pull/14913
- All Businesses logic for when the plugin should be shown works as expected when using `go run ./cmd/generator/ add mattermost-plugin-mscalendar v1.0.0 --official --enterprise`
- When tested along with the corresponding server PR the option to install the plugin is functional 
- See https://github.com/mattermost/mattermost-server/pull/14913#issuecomment-651381811 for testing details
LGTM! 

@hanzei I believe this ☝️  validates this market place change. Though I'm not sure when this should be merged. My concern is that once this is merged, users with 5.24 AND E20 will still see the broken button as they don't have this PR https://github.com/mattermost/mattermost-server/pull/14913
Thoughts? Can we preserve the existing market place functionality for 5.24 server?

hanzei:
@DHaussermann All 5.24 users should be able to see and install the plugin. This is covered by Test 2.-4.  https://github.com/mattermost/mattermost-marketplace/issues/77. Could you please double check that this is the case?

DHaussermann:
@hanzei just a heads-up that I don't see this working on out 5.25 test servers.  MS Calendar is always viable even without E20 license. The corresponding server PR was cherry picked 10 days ago. https://github.com/mattermost/mattermost-server/pull/14913 Not sure which repo the problem is in but I'll create a Jira ticket  for this.

Before I create a ticket - Is it possible the behavior won't work as expected without this PR? https://github.com/mattermost/mattermost-marketplace/pull/86

hanzei:
@DHaussermann Could my comment at https://github.com/mattermost/mattermost-plugin-github/pull/315#issuecomment-656850613 also resolve this question? I suspect you tested this changes with `v5.25.0-rc1`, not `v5.25.0-rc2`.

DHaussermann:
@hanzei  No I don't believe so. Currently 5.25.0-rc2 is deployed and I still see the plugin with no license in place. 

I do see a commit here https://github.com/mattermost/mattermost-marketplace/pull/86 that I can test along side 5.25-rc2 and will let you know.

PR #83, opened 2020-06-30
Add v1.4.0 of mattermost-plugin-zoom to the Marketplace

larkox:
#### Summary
Add Zoom v1.4.0 to Marketplace

#### Ticket Link
None


DHaussermann:
@larkox I have an issue with this. When upgrading via the marketplace the app functions as though JWT is set for Authentication which is the correct behavior.
But the UI appears as though neither option is set.
Screen Shot 2020-06-30 at 12 11 59 PM

This is not the behavior I saw in version bump testing (possibly having pre-existing data in my .json config hide the issue.) Is this something that can easily be resolved?

Steps:
- Spin up a new sever
- Install Zoom `v1.3.1` from production marketplace
- Configure JWT and enable
- Set marketplace to point somewhere that this PR is deployed
- Upgarde to Zoom 1.4
**Observed** Password based oAuth radio (which did not exist before) Does not show as being set to True.

larkox:
I would have to dig deeper into the server code to understand what it is doing.

My hope on this was that since "Enable Password based authentication" field was not set, it would take the default (true). But it seems that since there is some information, it just take the "0 value" of that field, in this case, false.

Not sure how to work around this that would be usable for the final user and maintenable in code.

DHaussermann:
@aaronrothschild Can you provide any input on if we should make an issue and investigate cutting a 1.4.1 release?

aaronrothschild:
We don't want to confuse users - this is the area that we want to be clear for the asmins when configuring things.  Having neither option show up isn't a huge deal, but we should fix asap.

DHaussermann:
Create https://github.com/mattermost/mattermost-plugin-zoom/issues/141 to address this.

larkox:
Due to the problem with the icon I understand we will not be adding v1.4.0 to the marketplace. Therefore closing this PR.

PR #89, opened 2020-07-13
Fix production

lieut-data:
#### Summary
`sls` wasn't working on `master`, so I've fixed to use the `serverless` name instead. Confirmed a successful deploy via `master`.

The commit history here is noisy, since I'm basically just merging `master` to `production`, and the merge commit there wasn't part of the tree that Git was expecting. The actual delta of code, however, is up-to-date.


ali-farooq0:
sls / serverless is a aws-cli thing?

ali-farooq0:
nvm just found this: https://www.serverless.com/framework/docs/providers/aws/cli-reference/deploy/ 👍

lieut-data:
Skipping QA review, since this has already been reviewed and only affects deployment scripts.

PR #90, opened 2020-07-15
Add v1.4.1 of mattermost-plugin-zoom to Marketplace

larkox:
#### Summary
This PR adds v1.4.1 of Zoom to marketplace

#### Ticket Link
None


larkox:
@jfrerich For some reason when I try to run "make generate", the statik file does not get updated correctly, creating this problem on the CI. Would you mind running yourself the "make generate" and see if it passes the CI?

plugins.json
```diff
@@ -4195,6 +4195,234 @@
...
+    "download_url": "https://github.com/mattermost/mattermost-plugin-zoom/releases/download/v1.4.1/zoom-1.4.1.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-zoom/releases/tag/v1.4.1",
+    "signature": "",
+    "manifest": {
+      "id": "zoom",
+      "name": "Zoom",
+      "description": "Zoom audio and video conferencing plugin for Mattermost 5.2+.",
```
aaronrothschild, line 4207:
```suggestion
      "description": "Zoom audio and video conferencing plugin for Mattermost 5.12+",
```

jfrerich:
@larkox, I pushed a commit with the results of `make generate`.

Question: Why is `v1.4.0` included in this release?  Was it not included before?

larkox:
The previous release was not released to the marketplace because it had a bug that we didn't want to show on the marketplace. Not sure if we should remove it from the marketplace, and if there is any "good way" to do that. Right now, if we were not to add it, we would have to remove it any time we do `make plugin.json` in the future.

jfrerich:
Thanks for the clarification.  I'm not sure we even allow getting a specific version from the marketplace.  Don't we only show the latest version anyway and in that case, we should be ok.

plugins.json
larkox: This is something that comes from the manifest, and it is not easily changed (we would need to cut a new release if I am not mistaken). Nevertheless, this is just on the description, so should not give big problems.

larkox:
I just realized that the signature for this release was never generated. I will look into the problem.

DHaussermann:
@larkox and @jfrerich when I deploy this to my local market place server. I see that `v1.4.1` is available. But, the upgrade does not work.
I see `Unable to verify plugin signature.` when I click Update. 
Screen Shot 2020-07-17 at 9 48 15 AM

DHaussermann:
Sorry @larkox I did not read the last comment before posting 🤦

jfrerich:
@larkox, if you need an assistance, feel free to DM and zoom me!

larkox:
@hanzei Can you take a look at this? For some reason, when I do `make generate` it does not create the right file. If you could create the right file and push the changes, I would really appreciate that.

hanzei:
@DHaussermann @iomodo @jfrerich @larkox I've run `make generate` to fix the build issue. LTAL

larkox:
@DHaussermann This should be ready to QA now.

Issue #74, opened 2020-06-04
Document `generator add`


hanzei:
https://github.com/mattermost/mattermost-marketplace/blob/master/cmd/generator/add.go is not documented. Add documentation.

PR #92, opened 2020-07-22
Add v0.3.0 of mattermost-plugin-todo to Marketplace

larkox:
#### Summary
This PR adds v0.3.0 of Todo to the Marketplace
#### Ticket Link
None


cmd/generator/main.go
```diff
@@ -96,6 +96,7 @@ var generatorCmd = &cobra.Command{
 			"mattermost-plugin-gitlab",
 			"mattermost-plugin-nps",
 			"mattermost-plugin-webex",
+			"mattermost-plugin-todo",
```
hanzei, line 99: The list is actually deprecated and should not be used any longer. It just have caused to much conflicts when multiple plugins get released at the same time. I will file a ticket to remove it.

larkox: Should I just remove that line? If I were to add other plugin without adding a line there, how would I do it? Just running `make plugin.json` did not add todo.

hanzei: Yes, that line should be removed.

You should run `go run ./cmd/generator/ add mattermost-plugin-todo v0.3.0 --beta` to add it.

I've filled https://github.com/mattermost/mattermost-marketplace/issues/74 to document this process.

plugins.json
```diff
@@ -1685,7 +1727,47 @@
...
     },
     "enterprise": false,
-    "updated_at": "2020-04-22T18:12:27.541293964Z"
+    "updated_at": "2020-04-21T17:20:23Z"
+  },
+  {
+    "homepage_url": "https://github.com/mattermost/mattermost-plugin-todo",
+    "icon_data": "",
+    "download_url": "https://github.com/mattermost/mattermost-plugin-todo/releases/download/v0.1.0/com.mattermost.plugin-todo-0.1.0.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-todo/releases/tag/v0.1.0",
```
hanzei, line 1736: Why should this release get added? The marketplace modal will only show the latest release anyway.

plugins.json
```diff
@@ -1648,10 +1648,52 @@
...
+        "header": "",
+        "footer": "",
+        "settings": []
+      }
+    },
+    "enterprise": false,
+    "updated_at": "2020-07-17T08:44:02Z"
+  },
+  {
+    "homepage_url": "https://github.com/mattermost/mattermost-plugin-todo",
```
hanzei, line 1693: Was this line added manually? If so, why?

larkox: After doing `make plugins.json` this was generated as it is now. Nothing added manually.

plugins.json
larkox: No special reason. My thought was that it will be added either way after each `make plugins.json` (since I added todo to the default repos), so I prefer to add it now instead of remove it in this pr and any other update on the list of plugins.

hanzei: Could you please only add `v0.3.0` in this PR? I would like to keep the diff as small as possible and avoid changes to the existing releases. Please me know if you need help.

mickmister:
Lint is saying

```
ERROR: uncommitted changes
 M data/statik/statik.go
```

Does the `statik.go` file need to be regenerated?

larkox:
@mickmister yes, it needs to be regenerated. For some reason, the statik.go file generated by my computer is not valid.

plugins.json
```diff
@@ -1647,6 +1647,48 @@
     "enterprise": false,
     "updated_at": "2019-06-12T19:04:44Z"
   },
+  {
+    "homepage_url": "https://github.com/mattermost/mattermost-plugin-todo",
+    "icon_data": "",
+    "download_url": "https://github.com/mattermost/mattermost-plugin-todo/releases/download/v0.3.0/com.mattermost.plugin-todo-0.3.0.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-todo/releases/tag/v0.3.0",
```
hanzei, line 1654: When I run `go run ./cmd/generator/ add mattermost-plugin-todo v0.3.0 --beta --official` locally, I get a different outcome for these lines:
```suggestion
    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-todo-v0.3.0.tar.gz",
    "release_notes_url": "",
```

Did you generate all of these changes using the same command as above? I curious to find out why my generated file is difference then yours.

plugins.json
```diff
@@ -1647,6 +1647,48 @@
...
+        "bundle_path": "webapp/dist/main.js"
+      },
+      "settings_schema": {
+        "header": "",
+        "footer": "",
+        "settings": []
+      }
+    },
+    "enterprise": false,
+    "updated_at": "2020-07-17T08:44:02Z"
```
hanzei, line 1690: Similar as above I'm getting a different line here 
```suggestion
    "updated_at": "2020-07-31T12:07:25.238392777Z"
```

As you can see it also includes (milli) seconds in the time stamp. Could that be the cause of the `statik.go` issue you are encountering?

PR #110, opened 2020-08-21
V2 Spec RFC

levb:


app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+    + [Lifecycle - Install](#lifecycle---install)
+    + [Lifecycle Uninstall.](#lifecycle-uninstall)
+    + [Configuration.](#configuration)
+    + [mattermost-cloud-app.json](#mattermost-cloud-appjson)
+  * [mattermost-plugin-cloud-apps](#mattermost-plugin-cloud-apps)
+  * [mattermost-server](#mattermost-server)
+    + [Minimal Changes to the Core Code.](#minimal-changes-to-the-core-code)
+  * [Cloud App Marketplace](#cloud-app-marketplace)
+
+## Cloud Apps.
```
iomodo, line 24: Do we consider plugins, custom slash commands, webhooks and bot accounts as Apps or are those separate entities?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Authentication with Mattermost.
+1. OAuth2 will be used as a primary mechanism of authentication for the Mattermost REST API.
```
iomodo, line 27: you probably meant authorization

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+2. Installing a **Cloud App** onto a Mattermost instance will provision sufficient credential to the app to invoke the REST API as the bot (priviledged? scoped? #TODO), and to verify/decode the JWT on the incoming requests from the Mattermost Server
+3. We should enable 3 use-cases:
+  a. Cloud App** using a "bot"-level token (to use APIs as the bot)
+  b. **Cloud App** using user-level OAuth2 tokens. Will require users' authoriation when first using the app.
```
iomodo, line 31: OAuth2 token will grant an app right to use user's data, act as a user, etc. Do we want to make it granular? For example "app may post only in the specific channels or DMs", "app gets access only to my public posts". Basically how detailed will be the scheme here?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+    + [Minimal Changes to the Core Code.](#minimal-changes-to-the-core-code)
+  * [Cloud App Marketplace](#cloud-app-marketplace)
+
+## Cloud Apps.
+
+### Authentication with Mattermost.
+1. OAuth2 will be used as a primary mechanism of authentication for the Mattermost REST API.
+2. Installing a **Cloud App** onto a Mattermost instance will provision sufficient credential to the app to invoke the REST API as the bot (priviledged? scoped? #TODO), and to verify/decode the JWT on the incoming requests from the Mattermost Server
+3. We should enable 3 use-cases:
+  a. Cloud App** using a "bot"-level token (to use APIs as the bot)
```
iomodo, line 30: Can you clarify what bot-level token means? What data/API can app access?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+  - UserLeftTeam
+  - UserUpdated
+2. Expandable context provides functions with most data dependencies pre-fetched, as instructed by the App Manifest, or the function binding.
+3. Overall scope in the App Manifest. #STRETCH individual scoped subscriptions for performance - channel, user, etc.
+
+### Autocomplete lookups
+- #TODO
+
+### Lifecycle - Install
+- Installation is initiated in the App Marketplace, from the Mattermost server. An application secret is generated by the App and must be provided to the Mattermost server.
```
larkox, line 116: Do we want to allow installation from outside mattermost? Use case:
- I am looking at tool A on their website.
- I go to their integrations site.
- I see they have a integration with Mattermost.
- I click the "install on mattermost" button they have on their website, and add my mattermost instance URL.
- It gets installed directly, without any need to look for it in the app marketplace, or the app being in the marketplace at all.

That could be controlled from the admin settings like "only allow apps from the marketplace".

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+## Cloud Apps.
+
+### Authentication with Mattermost.
+1. OAuth2 will be used as a primary mechanism of authentication for the Mattermost REST API.
+2. Installing a **Cloud App** onto a Mattermost instance will provision sufficient credential to the app to invoke the REST API as the bot (priviledged? scoped? #TODO), and to verify/decode the JWT on the incoming requests from the Mattermost Server
+3. We should enable 3 use-cases:
+  a. Cloud App** using a "bot"-level token (to use APIs as the bot)
+  b. **Cloud App** using user-level OAuth2 tokens. Will require users' authoriation when first using the app.
+  c. "A trusted app" - all users will automatically get "on-behalf" tokens, like with Atlassian Connect. @crspeller mentioned a "trusted app" concept in OAuth2, need to research.
+4. Function requests, and Event notifications sent to the Cloud App will include a JWT. The JWT will contain the acting Mattermost UserID (or Bot UserID for non-user-specific notifications)
```
iomodo, line 33: Will we have extra payload or claims besides UserID?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+  - UserLeftChannel
+  - UserLeftTeam
+  - UserUpdated
+2. Expandable context provides functions with most data dependencies pre-fetched, as instructed by the App Manifest, or the function binding.
+3. Overall scope in the App Manifest. #STRETCH individual scoped subscriptions for performance - channel, user, etc.
+
+### Autocomplete lookups
+- #TODO
+
+### Lifecycle - Install
```
iomodo, line 115: Can we say more about the lifecycle?
I assume there will be apps controlled by us, installed in the cluster, etc.
And there will be third-party apps, which will be running somewhere else. Correct?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+- Installation is initiated in the App Marketplace, from the Mattermost server. An application secret is generated by the App and must be provided to the Mattermost server.
+- Mattermost server completes the handshake with the app by sending it "App Installed" message, which includes the App secret. 
+- Once the handshake is complete, the `install` function is called on the App, as configured in the App Manifest
+
+### Lifecycle Uninstall.
```
iomodo, line 120: Will we determine which apps are running? For example which slash commands can even be called? stuff like that?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+### Mattermost to Cloud App.
+Mattermost interacts with Cloud Apps in 3 ways: 
+  - **Functions** that are invoked via Interactive Inputs, and have
+    request/response semantics. They may be synchronous, or execute in the
+    background once invoked.
+  - 1-way **Change notifications** that are sent after changes occur (from the
+    server).
+  - **Autocomplete** queries.
+
+### Functions
```
iomodo, line 64: Are functions registered by apps? What's the difference between function execution and calling an HTTP endpoint?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+### Interactive Inputs.
+See [Interactive Inputs](commands.md) for more.
+  - **Slash Commands** with autocomplete, execute functions.
+  - **Widgets** that can be bound to directly execute functions (Post menu items, channel menu, interactive post buttons, etc). In the future they may be extended to become in-place forms or autocomplete boxes.
+  - **Modal Dialogs** and **Bot Dialogs**, that gather/edit parameter values, then execute a function.
+
+#### Slash Commands
+1. Slash commands will have interactive auto-complete, available in the usual message box in Mattermost. All declared commands that are not "hidden" will be available for autocomplete.
+2. #TODO we need to be able to re-define what is visible (valid?) based on the state already entered, and hide the irrelevant options. @iomodo can you take a look at this?
+3. #STRETCH a "slash command" widget that would execute a pre-configured autocomplete command, in the given context (team/channel or Bot DM), and display the result.
```
iomodo, line 83: Can you clarify this?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+4. Slash commands may have a --modal option that would launch a modal to gather the inputs, rather than autocomplete. The parameters in the slash command are used as the default values in the modal.
+5. All valid slash commands must be declared in the App manifest
```
iomodo, line 85: Can slash commands be dynamically added/modified by the App?

app-marketplace-v2-spec/architecture.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Widgets
```
iomodo, line 87: How are widgets viewed in the MM?

app-marketplace-v2-spec/introduction.md
```diff
@@ -0,0 +1,96 @@
...
+   and review.
+4. **Plugin** installation and configuration management UX is neither
+   consolidated, nor consistent.
+5. **Plugins** are not sufficiently isolated neither in terms of data access,
+   nor in terms of performance for cloud (and some enterprise) deployments.
+6. **Slack** compatibility has not been actively maintained, is behind.
+
+## Project Summary 
+
+1. Build the **App Marketplace**, the place to discover,
```
larkox, line 35: If App Marketplace is going to be the place to manage plugins... does it need also capabilities to install and configure custom plugins? (i.e. plugins manually uploaded to the instance).

app-marketplace-v2-spec/structs.go
```diff
@@ -0,0 +1,125 @@
...
+	Description string
+	SiteURL     string
+
+	CommandHandlers []Function
+
+	SlashCommands []SlashCommand
+	ModalDialogs  []ModalDialog
+	BotDialogs    []BotDialog
+	Widgets       []Widget
+
```
larkox, line 14: I am starting to like having separated the functions from the widgets. We call define the functions by an internal name, and the endpoint they should reach. Then a widget is defined by type/location (whichever name we finally use), callable type (whether it runs a function or opens a dialog), and the name of the callable. The arguments passed will be the relevant on the widget context. Going down to manifest definition...

```
type Function struct {
    Name string  // e.g. create_meeting
    URL string  // e.g. https://myservice.com/start
    // Maybe extra static args we want to pass either way? Not sure how useful
    Args []Arg // e.g. {source: mattermost}
}

type Widget struct {
    Location string // e.g. CHANNEL_HEADER
    CallableType string // e.g. FUNCTION
    Callable string // e.g. create_meeting
    Extra interface{} // location dependant extra information (icon, text...)
}
```

Client will not receive the functions, and when a widget is activated, it just tells the server "run command X from app Y with these arguments".

levb: `Name` is the URL (with some common prefix). `CallableType` - maybe, I suppose would need it to decode even in go.

app-marketplace-v2-spec/introduction.md
levb: 0/5 - not yet. I propose that we keep all existing mechanisms in place, deprecate/consolidate them as a next "major release".

app-marketplace-v2-spec/architecture.md
levb: Still #TODO, but what's shaping up is that a "widget" really is a binding of a function call to a location, with args pre-specified. The component at the Location may have state (on/off checkbox, select, etc.).

app-marketplace-v2-spec/architecture.md
levb: No, but the "hidden" and "required" can be updated mid-entry. Just like the Modal Dialogs. Eventually we can add REST API calls to modify the definitions.

app-marketplace-v2-spec/architecture.md
levb: The 3 types of HTTP points exposed by the app are ^^^ (functions, notifications, lookups).

app-marketplace-v2-spec/structs.go
larkox: My point of giving them names is to make it more readable (and writable) the manifest.
e.g., I have registered the function create meeting, which is called by the channel header widget, and the start slash command.

The way I see it, the alternative would be something like:
I have registered the function that triggers the url `http://myservice.com/start`, and a widget that calls the function `http://myservice.com/start` and a slash command that calls the function `http://myservice.com/start`.

mickmister:
@levb Is this made obsolete by the google doc, or is this slightly different?

levb:
Just keeping it around until I assure that all comments are reflected in the Google Doc, but for all effective purposes this PR obsolete, will close shortly

PR #111, opened 2020-08-24
Add v1.2.0 of mattermost-plugin-gitlab to the Marketplace

iomodo:


hanzei (changes requested):
The base branch should be `production`, right?

PR #116, opened 2020-09-11
[MM-27942] Add platform-specific bundles if available

mickmister:
#### Summary

In order to cut down on the time it takes to download plugins on server start-up, the server can request the bundle for its specific architecture when installing from the marketplace. Then we can conditionally serve the arch-specific bundle if it exists.

The `download_url` and `signature` are edited in-place as we create the response. Discussion on this solution is in the comments of the linked ticket.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-27942

#### Related Pull Requests

https://github.com/mattermost/mattermost-server/pull/15436


internal/model/plugin_test.go
```diff
@@ -73,6 +73,12 @@ func TestPluginsFromReader(t *testing.T) {
 				Signature:       "signature1",
 				ReleaseNotesURL: "https://github.com/mattermost/mattermost-plugin-demo/releases/v0.1.0",
 				Manifest:        &mattermostModel.Manifest{},
+				ArchBundles: ArchBundles{
+					LinuxAmd64: &ArchBundleMetadata{
```
lieut-data, line 77: High-level, but would we consider a `LinuxDarwinWindowsAmd64` entry and make the the `Signature` and `DownloadURL` homogeneous?

ali-farooq0 (changes requested):
Overall looks good. Just at a high level though, I see we use a lot of `arch` keywords, lets either use `architecture` or if that is too long then maybe consider using another term for it, how about `platform`?

because someone new reading `plugin.ArchBundles = model.ArchBundles{}`, will definitely cause confusion, unless they know the context. I'd rather we're verbose and clear than concise and confusing

internal/model/plugin_test.go
mickmister: @lieut-data I'm not sure I understand. Do you mind showing me via an inline struct here?

mickmister:
@ali-farooq0 `platform` sounds good to me.

> someone new reading `plugin.ArchBundles = model.ArchBundles{}`, will definitely cause confusion

One question about this. Would this cause any issues for readability? `plugin.PlatformBundles = model.PlatformBundles{}`

internal/model/plugin_test.go
mickmister: @lieut-data and I discussed offline, and decided the proposed change above is not required, and we will go forward using the existing `DownloadURL` and `Signature` for the default bundles.

lieut-data (approved):

cmd/generator/bundles.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+// addPlatformSpecificBundles includes the platform-specific bundle URLs and signatures in the Marketplace entries.
+func addPlatformSpecificBundles(plugin *model.Plugin) (*model.Plugin, error) {
+	if plugin.HomepageURL == "" {
+		return plugin, nil
+	}
+
+	slashIndex := strings.LastIndex(plugin.HomepageURL, "/")
+	repo := plugin.HomepageURL[slashIndex+1:]
```
hanzei, line 21: E.g. MSCalender uses not the GitHub page as a `HomepageURL` but instead uses a GitBook URL. Can `addPlatformSpecificBundles` just accept a repo and org?

cmd/generator/bundles.go
```diff
@@ -0,0 +1,101 @@
...
+	platforms := []string{model.LinuxAmd64, model.DarwinAmd64, model.WindowsAmd64}
+	for _, platform := range platforms {
+		path := fmt.Sprintf("%s/%s-%s.tar.gz", remotePluginHost, pluginWithVersion, platform)
+
+		// Check if plugin bundle exists on remote file server
+		res, err := http.Head(path)
+		if err != nil {
+			return nil, err
+		}
+		if res.StatusCode != http.StatusOK {
```
hanzei, line 84: Worth logging a warning in this case?

cmd/generator/main.go
```diff
@@ -142,12 +145,17 @@ var generatorCmd = &cobra.Command{
 func getReleasePlugins(ctx context.Context, client *github.Client, repositoryName string, includePreRelease bool, existingPlugins []*model.Plugin) ([]*model.Plugin, error) {
 	logger := logger.WithField("repository", repositoryName)
 
-	repository, _, err := client.Repositories.Get(ctx, "mattermost", repositoryName)
+	org := os.Getenv("GITHUB_ORG")
+	if org == "" {
+		org = defaultGitHubOrg
+	}
```
hanzei, line 151: Is there a need to support a different org right now? If so, separate PR?

internal/model/plugin.go
```diff
@@ -19,8 +19,27 @@ type Plugin struct {
 	Manifest        *mattermostModel.Manifest `json:"manifest"`
 	Enterprise      bool                      `json:"enterprise"` // Indicated if the plugin is an enterprise plugin
 	UpdatedAt       time.Time                 `json:"updated_at"` // The point in time this release of the plugin was added to the Plugin Marketplace
+	PlatformBundles PlatformBundles           `json:"platform_bundles"`
```
hanzei, line 22: I'm would like 2/5 to see `UpdatedAt` being kept as the last field just to reduce the diff

hanzei (changes requested):
When I run `make plugins.json` locally, `arch_bundles` only contains `linux-amd64` and `windows-amd64` Could you check what is going on with `darwin-amd64`?

cmd/generator/main.go
mickmister: @hanzei I used this to test with my own GitHub org, because I thought I needed a GitHub access token for the mattermost org. But I now realize that you just need any GitHub access token so you can query GitHub in general.

ali-farooq0:
> @ali-farooq0 `platform` sounds good to me.
> 
> > someone new reading `plugin.ArchBundles = model.ArchBundles{}`, will definitely cause confusion
> 
> One question about this. Would this cause any issues for readability? `plugin.PlatformBundles = model.PlatformBundles{}`

Personally think that's fine. In the other case I was more concerned about the `Arch` part :D

But if you wanted to shorten it even further, it could become `plugin.Platforms` and with that the reader will probably be able to infer that these are other platform specific bundles. But maybe i'm reading too much into it. 🤷‍♂️  0/5 on this though.

ali-farooq0 (approved):
Looks good at a high level. Will pass the baton to @hanzei for the rest of it :)

cmd/generator/main.go
hanzei: I'm not against make the org configurable. But if you add this feature to this PR, we have to ensure it's available everywhere. Maybe even use a flag per command?

cmd/generator/bundles.go
mickmister: @hanzei I was using this to figure out the repo name, which we don't have a source of truth for in `plugins.json` at the moment. I am adding a `RepoName` field to the plugin entry so we have it available. We can backfill these in of course.

The plugins are being stored in s3 as their repo name, as opposed to their plugin id. For example:

`com.mattermost.mscalendar` is being served at `https://plugins-store.test.mattermost.com/release/mattermost-plugin-mscalendar-v1.0.0.tar.gz`.

So the raw repo name is what is needed to calculate this URL. I suggest storing it in the top level of the marketplace entries so the manifests don't need to be updated for each existing plugin, though we could start putting it there as well.

cmd/generator/main.go
mickmister: To decrease scope of this change, I will take the configurability out. 0/5 I'm not even sure it's worth making configurable in the first place for the purposes of the marketplace.

mickmister:
@hanzei I've responded to your comments above.

> When I run `make plugins.json` locally, `arch_bundles` only contains `linux-amd64` and `windows-amd64` Could you check what is going on with `darwin-amd64`?

@hanzei The reason for this is the s3 keys contain `osx-amd64`, where as the `runtime.GOOS` value is `darwin`:

```json
"darwin-amd64": {
	"download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-github-v0.14.0-osx-amd64.tar.gz",
}
```

I'm changing it to look for `osx` in the URL instead of `darwin`. @ali-farooq0 Do you know the history of why `osx-amd64` is used in the `s3` key rather than `darwin-amd64`?

plugins.json
```diff
@@ -4193,9 +5053,269 @@
...
+    "icon_data": "data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iOTEyIiBoZWlnaHQ9IjkxMiIgdmlld0JveD0iMCAwIDkxMiA5MTIiIGZpbGw9Im5vbmUiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+CjxjaXJjbGUgY3g9IjQ1NiIgY3k9IjQ1NiIgcj0iNDU2IiBmaWxsPSIjMkQ4Q0ZGIi8+CjxwYXRoIGQ9Ik0xOTkuNSAzNDkuMTI1QzE5OS41IDMzMy4zODUgMjEyLjI2IDMyMC42MjUgMjI4IDMyMC42MjVINDcwLjI1QzUyOS4yNzUgMzIwLjYyNSA1NzcuMTI1IDM2OC40NzUgNTc3LjEyNSA0MjcuNVY1NjIuODc1QzU3Ny4xMjUgNTc4LjYxNSA1NjQuMzY1IDU5MS4zNzUgNTQ4LjYyNSA1OTEuMzc1SDMwNi4zNzVDMjQ3LjM1IDU5MS4zNzUgMTk5LjUgNTQzLjUyNSAxOTkuNSA0ODQuNVYzNDkuMTI1WiIgZmlsbD0id2hpdGUiLz4KPHBhdGggZmlsbC1ydWxlPSJldmVub2RkIiBjbGlwLXJ1bGU9ImV2ZW5vZGQiIGQ9Ik02NzYuODE1IDU4My44MjlMNjE4LjI5MyA1MzguNjUzQzYxNC43OTcgNTM1Ljk1NSA2MTIuNzUgNTMxLjc4OSA2MTIuNzUgNTI3LjM3M1Y0MTMuMTI3QzYxMi43NSA0MDguNzExIDYxNC43OTcgNDA0LjU0NSA2MTguMjkzIDQwMS44NDdMNjc2LjgxNSAzNTYuNjcxQzY4MS4zNDUgMzUyLjAxNiA2ODcuNjc5IDM0OS4xMjUgNjk0LjY4OCAzNDkuMTI1QzcwOC40NiAzNDkuMTI1IDcxOS42MjUgMzYwLjI5IDcxOS42MjUgMzc0LjA2MlY1NjYuNDM4QzcxOS42MjUgNTgwLjIxIDcwOC40NiA1OTEuMzc1IDY5NC42ODggNTkxLjM3NUM2ODcuNjc5IDU5MS4zNzUgNjgxLjM0NSA1ODguNDg0IDY3Ni44MTUgNTgzLjgyOVoiIGZpbGw9IndoaXRlIi8+Cjwvc3ZnPgo=",
```
mickmister, line 5062: It looks like there are some new entries here after running `make plugins.json`. @hanzei Any idea why this might be happening? For the purposes of this PR, I suggest we just update `repo_name` and the `platforms` properties and do not add any "missing" entries.

mickmister: I actually decided to include them, to avoid doing manual surgery on `plugins.json`

mickmister: The entries that were added:

- [gitlab 1.2.0](https://github.com/mattermost/mattermost-plugin-gitlab) (latest)
- [nps 1.0.2](https://github.com/mattermost/mattermost-plugin-nps) (not latest)
- [aws SNS 1.1.0](https://github.com/mattermost/mattermost-plugin-aws-SNS) (latest)
- [custom attributes 1.3.0](https://github.com/mattermost/mattermost-plugin-custom-attributes) (latest)
- [jira 3.0.0](https://github.com/mattermost/mattermost-plugin-jira) (latest)
- [autolink 1.2.1](https://github.com/mattermost/mattermost-plugin-autolink) (latest)
- [zoom 1.4.1](https://github.com/mattermost/mattermost-plugin-zoom) (latest)
- [zoom 1/4.1](https://github.com/mattermost/mattermost-plugin-zoom) (not latest)

mickmister: @jfrerich do you have any ideas why this may be happening when I run `make plugins.json`?

mickmister: I did the generate process over again, and made sure that any new unwanted entries were removed. I think this is fine for this PR, but I think we need to figure out why the new entries are being added.

internal/model/plugin.go
```diff
@@ -18,9 +19,29 @@ type Plugin struct {
...
+// PlatformBundleMetadata holds the necessary data to fetch and verify a plugin built for a specific platform
+type PlatformBundleMetadata struct {
+	DownloadURL string `json:"download_url"`
+	Signature   string `json:"signature"`
+}
+
+type PlatformBundles struct {
+	LinuxAmd64   *PlatformBundleMetadata `json:"linux-amd64,omitempty" yaml:"linux-amd64,omitempty"`
+	DarwinAmd64  *PlatformBundleMetadata `json:"darwin-amd64,omitempty" yaml:"darwin-amd64,omitempty"`
+	WindowsAmd64 *PlatformBundleMetadata `json:"windows-amd64,omitempty" yaml:"windows-amd64,omitempty"`
```
hanzei, line 35: Does this need to be a pointer? Can we simplify by using a struct directly?

cmd/generator/bundles.go
```diff
@@ -0,0 +1,96 @@
...
+			logger.Infof("Platform-specific bundle not found %s", path)
```
hanzei, line 77: Is it worth also logging the plugin id?

cmd/generator/main.go
```diff
@@ -27,11 +27,16 @@ import (
...
 
+const defaultRemotePluginHost = "https://plugins-store.test.mattermost.com/release"
+const defaultGitHubOrg = "mattermost"
+
 func init() {
 	generatorCmd.PersistentFlags().Bool("debug", false, "Whether to output debug logs.")
 	generatorCmd.PersistentFlags().String("database", "plugins.json", "Path to the plugins database to update.")
 
 	generatorCmd.Flags().Bool("include-pre-release", false, "Whether to include pre-release versions.")
+	generatorCmd.Flags().String("github-org", defaultGitHubOrg, "GitHub org that owns the plugin releases.")
```
hanzei, line 38:
```suggestion
	generatorCmd.Flags().String("github-org", defaultGitHubOrg, "GitHub organisation that owns the plugin releases.")
```

cmd/generator/main.go
```diff
@@ -104,11 +119,10 @@ var generatorCmd = &cobra.Command{
 			logger.Debugf("querying repository %s", repositoryName)
 
 			var releasePlugins []*model.Plugin
-			releasePlugins, err = getReleasePlugins(ctx, client, repositoryName, includePreRelease, existingPlugins)
+			releasePlugins, err = getReleasePlugins(ctx, client, githubOrg, repositoryName, pluginHost, includePreRelease, existingPlugins)
 			if err != nil {
 				return errors.Wrapf(err, "failed to release plugin for repository %s", repositoryName)
 			}
-
```
hanzei, line 111: Nit: unrelated change

internal/model/plugin.go
```diff
@@ -10,6 +10,7 @@ import (
 
 // Plugin represents a Mattermost plugin in the Plugin Marketplace.
 type Plugin struct {
+	RepoName        string                    `json:"repo_name"`
```
hanzei, line 13: 1/5 that this filed should be after `Signature` as `HomepageURL` and  `IconData` are directly from the manifest

internal/model/plugin.go
```diff
@@ -18,9 +19,29 @@ type Plugin struct {
...
+
+type PlatformBundles struct {
+	LinuxAmd64   *PlatformBundleMetadata `json:"linux-amd64,omitempty" yaml:"linux-amd64,omitempty"`
+	DarwinAmd64  *PlatformBundleMetadata `json:"darwin-amd64,omitempty" yaml:"darwin-amd64,omitempty"`
+	WindowsAmd64 *PlatformBundleMetadata `json:"windows-amd64,omitempty" yaml:"windows-amd64,omitempty"`
+}
+
+const (
+	LinuxAmd64   = "linux-amd64"
+	DarwinAmd64  = "darwin-amd64"
```
hanzei, line 40: Why is this forth type needed? Can we use `OsxAmd64` everywhere instead?

internal/store/static.go
```diff
@@ -140,6 +140,11 @@ func (store *StaticStore) getPlugins(serverVersion string, includeEnterprisePlug
 		}
 
 		if serverVersion != "" && storePlugin.Manifest.MinServerVersion != "" {
+			_, err := semver.Parse(serverVersion)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to parse serverVersion %s", serverVersion)
+			}
```
hanzei, line 146: Should this check get moved to after L123 to DRY the code?

internal/store/static.go
```diff
@@ -150,6 +155,28 @@ func (store *StaticStore) getPlugins(serverVersion string, includeEnterprisePlug
 			}
 		}
 
+		if platform != "" {
+			var bundle *model.PlatformBundleMetadata
+
+			// We don't want to edit the existing plugin's download_url. Only the response should be edited.
+			newRef := *storePlugin
```
hanzei, line 162: Could you elaborate a bit? What is the downside of editing the original variable `storePlugin`?

plugins.json
```diff
@@ -989,9 +1045,11 @@
       }
     },
     "enterprise": false,
+    "platforms": {},
```
hanzei, line 1048: That is surprising. There should be one for this plugin. Could you please check why there isn't?

plugins.json
```diff
@@ -1157,9 +1232,11 @@
...
```
hanzei, line 1235: Same here. There should be one.

plugins.json
```diff
@@ -1437,9 +1550,11 @@
       }
     },
     "enterprise": true,
+    "platforms": {},
```
hanzei, line 1553: And here

plugins.json
```diff
@@ -1685,9 +1836,11 @@
       }
     },
     "enterprise": false,
+    "platforms": {},
```
hanzei, line 1839: And here

plugins.json
```diff
@@ -4193,9 +4563,11 @@
...
```
hanzei, line 4566: And here

hanzei (changes requested):
A couple of plugins are missing the `platforms`. Could you please check why the assets were not found?

internal/model/plugin.go
mickmister: It was to make it simple to verify presence of the bundle. Are you implying we do this:
```suggestion
	LinuxAmd64   PlatformBundleMetadata `json:"linux-amd64,omitempty" yaml:"linux-amd64,omitempty"`
	DarwinAmd64  PlatformBundleMetadata `json:"darwin-amd64,omitempty" yaml:"darwin-amd64,omitempty"`
	WindowsAmd64 PlatformBundleMetadata `json:"windows-amd64,omitempty" yaml:"windows-amd64,omitempty"`
```

Then when checking for a given platform's presence, we do something like `if platforms.LinuxAmd64.URL != ""` rather than `if platforms.LinuxAmd64 != nil`?

internal/store/static.go
mickmister: I was mistakenly thinking this would be running ongoing with the plugin store in RAM, but it is running in a stateless lambda function, where any changes do not get committed back to the store. I will change it so it does not dereference.

internal/model/plugin.go
hanzei: Yes, that was my suggestion

internal/store/static.go
mickmister: I tried this but there is some logic of `serverVersion == ""` in both of these blocks, and the `semver.Parse` function throws an error when it is passed an empty string. So I'll keep it how it is unless you see another way to improve the flow.

mickmister:
@hanzei All requests have been addressed. Please see my response on the remaining unresolved comment regarding checking semver

internal/store/static.go
hanzei: Good point. It's probably fine for now.

internal/model/plugin.go
```diff
@@ -16,11 +16,31 @@ type Plugin struct {
...
 	Manifest        *mattermostModel.Manifest `json:"manifest"`
 	Enterprise      bool                      `json:"enterprise"` // Indicated if the plugin is an enterprise plugin
+	Platforms       PlatformBundles           `json:"platforms"`
 	UpdatedAt       time.Time                 `json:"updated_at"` // The point in time this release of the plugin was added to the Plugin Marketplace
 }
 
+// PlatformBundleMetadata holds the necessary data to fetch and verify a plugin built for a specific platform
+type PlatformBundleMetadata struct {
+	DownloadURL string `json:"download_url"`
+	Signature   string `json:"signature"`
```
hanzei, line 29: Is it worth setting these as `omitempty` to simplify `plugins.json`?

hanzei (approved):
Great work! A few comments and questions from my side but nothing blocking.

cmd/generator/bundles.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	pluginWithVersion := ""
+	if strings.HasPrefix(plugin.DownloadURL, pluginHost) {
+		pluginPath := strings.TrimPrefix(plugin.DownloadURL, pluginHost+"/")
+		pluginWithVersion = strings.TrimSuffix(pluginPath, ".tar.gz")
+	} else {
+		repo := plugin.RepoName
+		pluginWithVersion = fmt.Sprintf("%s-v%s", repo, plugin.Manifest.Version)
+	}
```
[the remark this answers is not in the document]
mickmister: There is one plugin entry where the repo name does not exactly match the URL structure. For the community-developed giphy plugin, here are the two values:

```json
{
    "repo_name": "mattermost-plugin-giphy",
    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-giphy-moussetc-v1.3.0.tar.gz"
}
```

Notice the difference between `mattermost-plugin-giphy` and `mattermost-plugin-giphy-moussetc`. This is the only reason why the first `if` block was added.

cmd/generator/bundles.go
```diff
@@ -0,0 +1,150 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/mattermost/mattermost-marketplace/internal/model"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+// OSX-specific bundle URLs are stored in the plugin store as `osx` rather than `darwin`
+const OsxAmd64 = "osx-amd64"
```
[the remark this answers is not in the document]
mickmister: This shows up in the next function at line 121. I'll keep it as-is to avoid duplication

cmd/generator/bundles.go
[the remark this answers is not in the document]
hanzei: In that case `repo_name` should be `mattermost-plugin-giphy-moussetc` as https://github.com/mattermost/mattermost-plugin-giphy-moussetc/ is the repository that contains the code.

Can we drop the `if` then?

cmd/generator/bundles.go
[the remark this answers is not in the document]
hanzei: Got it, missed that one :+1:

hanzei (approved):

hanzei (approved):

internal/store/static.go
mickmister: @hanzei After some testing, I've realized that I originally did this because the mutation persists in a local deployment. `plugins.json` must be kept in RAM between requests for the local deployment. I've changed it back to dereference.

mickmister:
@DHaussermann I've confirmed this works on all supported platforms: `Windows Server`, `MacOS`, `Linux`. What are your thoughts on pushing this PR through?

internal/store/static.go
```diff
@@ -150,6 +155,25 @@ func (store *StaticStore) getPlugins(serverVersion string, includeEnterprisePlug
...
+				bundle = storePlugin.Platforms.DarwinAmd64
+			case model.WindowsAmd64:
+				bundle = storePlugin.Platforms.WindowsAmd64
+			}
+
+			if bundle.DownloadURL != "" && bundle.Signature != "" {
+				newRef := *storePlugin
+				newRef.DownloadURL = bundle.DownloadURL
+				newRef.Signature = bundle.Signature
+				storePlugin = &newRef
```
hanzei, line 173: Can you please elaborate why the pointer shenanigans are needed?

mickmister: @hanzei I replied on the other conversation https://github.com/mattermost/mattermost-marketplace/pull/116#discussion_r510591763 

When the server runs locally, it caches the `plugins.json` contents in memory. When the pointer of the struct is edited here (without my hacky fix here), it edits the master copy that is stored in memory. Maybe we should just not use pointers to reference the source data.

mickmister: @hanzei If this is something that needs to be addressed, do you mind if I make it a follow up ticket so we can merge this?

I've tried the method of using `[]model.Plugin` instead of `[]*model.Plugin` to avoid mutation of the source data, but a lot of different structs here implement the interface that has the method's return value as `[]*model.Plugin`. If you think avoiding pointers altogether is the right solution, 2/5 I think we should offload this change to another ticket.

hanzei: Surely we can merge this PR as it is. I didn't intended to block QA review.

mickmister:
@DHaussermann what are your thoughts on my comment above? I'd like to get this to the finish line for the cloud instances. We can meet to discuss if you'd like

DHaussermann (approved):
Tested and passed
- Tested to ensure that download URL is now platform specific 
- Tested to ensure 3 unique versions of the download URL now serve only the binnary for the platform being used 
- Tested to ensure marketplace still serves all 3 binaries when in use on an older server version 
LGTM!

Will add Cypress tests for market place to cover this.

PR #128, opened 2020-10-09
[MM-29362] Allow filtering by hosting type

hanzei:
#### Summary
This PR adds a `hosting` field to the stored plugins. If it's set to either `cloud` or `on-prem` the plugin will be limited to that kind of installation.

The PR does not contain any changes to the current list of plugins. These will be added in a separate PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-29362

#### Related Pull Requests

https://github.com/mattermost/mattermost-server/pull/15913


cmd/generator/add.go
```diff
@@ -52,6 +54,20 @@ var addCmd = &cobra.Command{
...
+			return err
+		}
+
+		onPrem, err := command.Flags().GetBool("on-prem")
+		if err != nil {
+			return err
+		}
+
+		if cloud && onPrem {
+			return errors.New("if you want to make a plugin availed for cloud and on-prem, just drop both flags")
```
catalintomai, line 68:
```suggestion
			return errors.New("if you want to make a plugin available for cloud and on-prem, just drop both flags")
```

hanzei: Good catch :+1:

catalintomai (approved):

internal/api/plugins_test.go
```diff
@@ -395,6 +415,54 @@ func TestPlugins(t *testing.T) {
...
+
+			plugins, err := client.GetPlugins(&api.GetPluginsRequest{
+				PerPage: -1,
+				Cloud:   true,
+			})
+			require.NoError(t, err)
+			require.ElementsMatch(t, []*model.Plugin{plugin1V3Min515, plugin2V1Min516, plugin3V3Min517}, plugins)
+		})
+
+		t.Run("on-prem plugin is not return for on-prem instance", func(t *testing.T) {
```
mickmister, line 454:
```suggestion
		t.Run("on-prem only plugin is not return for cloud instance", func(t *testing.T) {
```

internal/model/plugin_test.go
```diff
@@ -156,6 +157,7 @@ func TestPluginsToWriter(t *testing.T) {
       "version": "2.0.0"
     },
     "enterprise": false,
+    "hosting": "",
```
mickmister, line 160: Does it make sense to have one test assertion here for the hosting being cloud or on-prem? As opposed to only having the empty string case.

mickmister (approved):
LGTM, just some non-blocking suggestions on tests

internal/model/plugin_test.go
hanzei: Makes sense. Added :+1:

internal/api/plugins_test.go
hanzei: Good catch :+1:

hanzei:
@DHaussermann I've added test steps to this in https://mattermost.atlassian.net/browse/MM-29362

DHaussermann (approved):
Tested and passed 
This is working as expected when I followed the test steps provided in the Jira task linked. We can now set where a plugin is `cloud` or `on-prem` in the marketplace config. 
LGTM!

Thanks @hanzei for steps and guidance on this one.

PR #131, opened 2020-10-21
Add v1.0.0 of mattermost-plugin-incident-response to the Marketplace

crspeller:
#### Summary
Upgrading the IR plugin in private beta to 1.0.


cpoile (approved):

plugins.json
```diff
@@ -31,6 +31,38 @@
...
+      "webapp": {
+        "bundle_path": "webapp/dist/main.js"
+      },
+      "settings_schema": {
+        "header": "",
+        "footer": "",
+        "settings": []
+      }
+    },
+    "enterprise": false,
```
itao, line 63: Does this mean it isn't listed with `E20` label?

crspeller:
@itao @cpoile Not sure what that does. The other versions if IR on the private beta store also have `"enterprise": false`. 
Checked the production store, the only one that has that set to true is the MS cal plugin.

crspeller:
@itao Ok so setting that to true will add the label, it will also limit the visibility of the plugin to only installations with an E20 license. If we are good with that on the private beta marketplace, then I will change it to true.

itao:
@crspeller Ok thanks, yea let's do that for both this private marketplace as well as the public one.

itao (approved):

hanzei:
I'm a bit late to the party: Should the IR plugin have an Enterprise Label similar to MS calendar?

itao:
@hanzei Yes it should and does with this PR :)

hanzei:
Actually, it wasn't added in this PR. Added in https://github.com/mattermost/mattermost-marketplace/pull/134.

PR #134, opened 2020-10-30
Add Enterprise label to Incident Response v1.0.0

hanzei:
#### Summary
Adding the Enterprise label was forgotten in https://github.com/mattermost/mattermost-marketplace/pull/131. This PR adds it.

#### Ticket Link
Follow up on https://github.com/mattermost/mattermost-marketplace/pull/131


crspeller (approved):
Looks like this should be automatic with that enterprise flag 🤦

PR #137, opened 2020-11-04
Programmatically add labels

hanzei:
#### Summary
Instead of storing all labels in the DB we now programmatically generate the list using metadata stored in the DB file. That ensure there is a single source of truth. The stored metadata will later also be used for filtering on the client side.

Given that `model.Plugin` in the store can't be send to the client any longer because the labels have to get added, I implemented a `AddLabels()` method that takes care of this. I'm unsure where the call it it belongs. Right now it's called after the plugins are read from store.

I thought about spiting `model.Plugin` into a two structs: A stored plugin and a returned plugin. E.g. `Labels` should not be stored and only returned and `Platforms` should not be returned via the API. But that seamed like to big change to piggy back into this PR. Still, I'm wondering if it's a valuable change.

Migration was done via the migration command. No manually edits.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-30240


internal/model/plugin.go
```diff
@@ -15,18 +15,34 @@ const (
...
 	HomepageURL     string                    `json:"homepage_url"`
 	IconData        string                    `json:"icon_data"` // The base64 encoding of an svg image
 	DownloadURL     string                    `json:"download_url"`
 	ReleaseNotesURL string                    `json:"release_notes_url"`
 	Labels          []Label                   `json:"labels,omitempty"`
-	Hosting         HostingType               `json:"hosting"`   // Indicated if the plugin is limited to a certain hosting type
-	Signature       string                    `json:"signature"` // A signature of a plugin saved in base64 encoding.
+	Hosting         HostingType               `json:"hosting"`    // Indicated if the plugin is limited to a certain hosting type
+	Maintainer      Maintainer                `json:"maintainer"` // The maintainer of the plugin
+	Stage           Stage                     `json:"stage"`      // The stage in the software release cycle that the plugin is in
```
hanzei, line 41: I'm 1/5 on the name `Maintainer` and 0/5 on `Stage`. Suggestions are welcome!

lieut-data:
@hanzei, can you elaborate on the rationale for making this change? I see "single source of truth", but I'm not sure what the "other" source of truth was in this case. Happy to continue reviewing, but just want to make sure I understand why we want to start modelling these properties directly instead of keeping them flexible as labels. Could we just filter client-side on the labels vs. hard-coding a particular schema that might have to change in the future?

lieut-data (approved):
I had a chance to sync up with @hanzei offline on the high-level questions -- changes look good!

PR #142, opened 2020-11-16
MM-28861 Add endpoint to marketplace to retrive all versions of specified plugin.

crspeller:
#### Summary
Adds the endpoint `/api/v1/plugin/{id}` to retrieve all the version of a specified plugin. Will still follow all filter rules except the text filter. 
Needed for server PR to give feature flags the ability to install specific versions of plugins.


hanzei (approved):
The code looks good to me. One thought below:

I'm wondering if `GetPlugin` should behave the same way as `GetPlugins` and return only the latest version. To accomplish the behaviour you want, we would add a new field `IncludePrevious` (0/5 on the same) into `PluginFilter` that would then return also older version. That way both `GetPlugins` and `GetPlugin` would support getting only the latest version and all versions.

internal/api/client.go
```diff
@@ -64,3 +64,26 @@ func (c *Client) GetPlugins(request *GetPluginsRequest) ([]*model.Plugin, error)
 		return nil, errors.Errorf("failed with status code %d", resp.StatusCode)
 	}
 }
+
+// GetPlugin fetches all the specified version of a single plugin
+func (c *Client) GetPlugin(request *GetPluginsRequest, pluginid string) ([]*model.Plugin, error) {
```
lieut-data, line 69: nit: `pluginId`?

internal/store/static.go
```diff
@@ -74,6 +74,58 @@ func pluginMatchesFilter(plugin *model.Plugin, filter string) bool {
 	return false
 }
 
+// GetPlugin retrieves all known version of a specific plugin compatable with the plugin filter given.
```
lieut-data, line 77:
```suggestion
// GetPlugin retrieves all known versions of a specific plugin compatible with the given plugin filter.
```

internal/store/static.go
```diff
@@ -74,6 +74,58 @@ func pluginMatchesFilter(plugin *model.Plugin, filter string) bool {
...
+	end := (pluginFilter.Page + 1) * pluginFilter.PerPage
+	if start >= len(plugins) {
+		return nil, nil
+	}
+	if end > len(plugins) {
+		end = len(plugins)
+	}
+
+	return plugins[start:end], nil
+
```
lieut-data, line 126: nit: spurious extra line

internal/store/static.go
```diff
@@ -74,6 +74,58 @@ func pluginMatchesFilter(plugin *model.Plugin, filter string) bool {
...
+	if pluginFilter.PerPage == 0 {
+		return nil, nil
+	}
+
+	plugins, err := store.getPlugins(pluginFilter.ServerVersion, pluginFilter.EnterprisePlugins, pluginFilter.Cloud, pluginFilter.Platform)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get plugins")
+	}
+
+	// Sort the final slice by plugin name, ascending
```
lieut-data, line 88:
```suggestion
	// Sort the final slice by plugin version ascending, then plugin name ascending.
```

lieut-data (changes requested):
Changes makes sense, but I do share @hanzei's sentiment about the significant semantic differences between `/plugins` and `/plugin`. For better or worse, the semantics-to-date have been "one version per plugin" returned, and there's a certain elegance to preserving that given the longevity of this service.

I do like @hanzei's suggestion of a flag to change the behaviour here (maybe `ReturnAllVersions`?). If we added `PluginId` as another flag, we could even just keep the current endpoint and be able to use the new `ReturnAllVersions` to dump everything if so desired. Open to suggestions here, though!

crspeller:
Sorry for the force push. Had to because of the strange branches. To properly test I had to rebase on to the production branch but then back to this one.

lieut-data (approved):

lieut-data:
One endpoint to rule them all!

PR #156, opened 2020-12-17
Add incident management v1.2.0.

lieut-data:
#### Summary
List the updated incident management v1.2.0 in the marketplace.


hanzei (dismissed):
Nice :+1: 

Do we need to update the prepackaed version? Or is there a feature flag already?

plugins.json
```diff
@@ -2705,6 +2705,66 @@
...
+      },
+      "windows-amd64": {
+        "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-incident-management-v1.2.0-windows-amd64.tar.gz",
+        "signature": "iQIzBAABCAAdFiEExViBuA9p6GO4WtXR0bVLR6XO/sQFAl/aeJYACgkQ0bVLR6XO/sStXQ/9HhBT8BQ19+H2BNhNrmlz7u5F64dflgIpY3180oUshkPXSS2QR43MbxW8dleUdTZgcBMq7uqVfuhAY56MKTJ8guzBDJlT6bo6PRUaIg53AfBd3cQKBl9ViW0QxX7MWP/T9oLATTTN64mUdRZrnH37oIHzymD/tDEqMzvokSZjWpCwflbyHCx2xXShzM1AnKAmlOQ04mVONFGsR0kvNxa9n1tjvxMk1YwrNHKQ7ybrTsjo3fYY9qg1MqFgX3HZRCiEEdT2Cs601tQgcmO2lqQ5TzQgn5/aKeVvb9733Shyd8A/l6AfKVniT5x1xhOUaQiLC6INlbwvoQJkD798UeUSEA0nC6C0oJ2pz25WGde6bYy1KBy6qNUEH/QzEPLmXbneP5VXfU8JA8NvJMdcm4L82+BJcGtbO13udeCkdC3aKkrM9cThLFfLQU81CciZ+Uoq8LIEa7ur0x3OgQ/2l1sPfbSMZ0tETXIg0IA+4puRqBZceb7ED4Ln+H/0q6bOn1EtH1KbzUtOJ2MZTVslWPOKSYbcI/UGGR8ltk8caI4CP4Nt07qgH6bu5FcLTNJhvipT9FzCBrH0X4Fx4/PMSSby3qd17gumF4Tcs9Yd/HKUexKBSVBnULZyTUQCQMdDfR2lQfziX6c2n8iWq4r8zxO/QJk5jzh6tEiPgr0fWIeZ3SM="
+      }
+    },
+    "updated_at": "2020-12-17T01:22:00.432331Z"
+  },
   {
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-incident-response/",
```
agarciamontoro, line 2769: I noticed that there is still an incident-response here. Is this intended?

agarciamontoro (dismissed):
Apart from the comment above, this PR looks good! :tada:

plugins.json
lieut-data: Yeah, I only just updated the url-related metadata for `plugin.json` in v1.2.0, so the old metadata for the older versions remains incorrect. Given the permanent redirect, I think we can leave this as-is, and it will gradually become less relevant as new versions are released.

lieut-data:
Thanks, @hanzei and @agarciamontoro! Yes, we do still want to pre-package this for v5.31 self-managed customers, but I believe we also have other options available to us for deploying more quickly. I don't think we'll be using any of those over the holidays, but will "deploy" in the new year as such.

lieut-data:
@prapti, once this gets merged, it will be "live", so I'll wait for an "all-clear" from your perspective re: v1.2.0.

prapti (approved):
Looks great! Tested locally.
Approving 🎉

hanzei (approved):

agarciamontoro (approved):
Ready to go!

PR #170, opened 2021-02-21
[MM-33116] Use embed package to include Marketplace plugin list for lambda

hanzei:
#### Summary
The generated file `data/statik/statik.go` has caused some issue in the past like unneeded merge conflicts and unpredictable in the file itself. Since go 1.16 files can be embed directly into the binary. By using that approach we can drop the generated file entirely and use that new mechanism.

We might want to wait a week or two before we merge this PR given that go 1.16 will be required then.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-33116


Makefile
```diff
@@ -9,17 +9,11 @@ LDFLAGS += -X "github.com/mattermost/mattermost-marketplace/internal/api.buildHa
 LDFLAGS += -X "main.upstreamURL=$(BUILD_UPSTREAM_URL)"
 SLS_STAGE ?= "dev"
 
+$(shell cp plugins.json ./cmd/lambda/)
```
hanzei, line 12: Given that there is no path traversal we need to copy the file into `cmd/lambda/`

larkox (approved):
LGTM!

lieut-data (approved):
This is terrific!

DHaussermann:
@hanzei Can you please clarify the scope or provide some guidance on testing for this change?

Is this something that would be covered by regression testing?

hanzei:
@DHaussermann The only production facing change of this PR is the loading of the listed plugins. Hence,  a quick smoke that if plugins are correctly listed is enough.

DHaussermann:
Thanks @hanzei I have regression tested the existing marketplace. I don't see any change in behavior. 

I also notice the http-hello app is listed and istallable for me but, these seems to have come from a previous PR correct?

image

hanzei:
@DHaussermann Correct. That is an example App provided by the Apps plugin.

DHaussermann (approved):
Tested and passed
Regression tested marketplace functionality.  No console errors or changes in behavior found.  
All UI actions are functional

LGTM!

PR #175, opened 2021-03-11
Add v3.1.1 of mattermost-plugin-dice-roller to the Marketplace

hanzei:
#### Summary

Screenshot from 2021-03-11 13-53-55

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-marketplace/issues/122


iomodo (approved):

DHaussermann (dismissed):
Tested and passed
- Plugin displays in marketplace as expected
- Community label is shown
- New icon displays as expected and looks good on dark theme
- Tested install and enable
- Version shows as 3.1.1

LGTM!

plugins.json
```diff
@@ -1168,6 +1168,63 @@
...
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-dice-roller-v3.1.1.tar.gz",
```
jespino, line 1174: Is this the valid download url?

hanzei: Yes

jespino (approved):

plugins.json
```diff
@@ -1168,6 +1168,63 @@
...
+        "executables": {
+          "linux-amd64": "server/dist/plugin-linux-amd64",
+          "darwin-amd64": "server/dist/plugin-darwin-amd64",
+          "windows-amd64": "server/dist/plugin-windows-amd64.exe"
+        },
+        "executable": "server/dist/plugin-freebsd-amd64"
+      },
+      "settings_schema": {
+        "header": "",
+        "footer": "* To report an issue, make a suggestion or a contribution, [check the GitHub repository]i(https://github.com/moussetc/mattermost-plugin-dice-roller/)",
```
agnivade, line 1207: Is the `i` in the middle intentional?

plugins.json
```diff
@@ -1168,6 +1168,63 @@
...
+      "icon_path": "assets/icon.svg",
+      "version": "3.1.1",
+      "min_server_version": "5.20.0",
+      "server": {
+        "executables": {
+          "linux-amd64": "server/dist/plugin-linux-amd64",
+          "darwin-amd64": "server/dist/plugin-darwin-amd64",
+          "windows-amd64": "server/dist/plugin-windows-amd64.exe"
+        },
+        "executable": "server/dist/plugin-freebsd-amd64"
```
agnivade, line 1203: Any reason freebsd-amd64 is treated separately than the executables above?

plugins.json
hanzei: You got eagle eyes! 

No, it's not intentionally. I only caught it after the plugin release was already done (but not pushed to the Marketplace, this is what this PR does). It's fixed upstream with 
 https://github.com/moussetc/mattermost-plugin-dice-roller/commit/a4597d56f873b09449c1e0a61b64fa3057786cf6 and the next release will contain the fix.

plugins.json
hanzei: Great question. The plugin framework only supports a mapping from `linux-amd64`, `darwin-amd64` and `windows-amd64` to a binary name. If the platform name of a given machine is neither of these three the binary in `executable` will be used. So this is a hacky way to support freeBSD.

Since recently `executables` is a generic `map[string]string` and hence all platforms can be listed there. This plugin just hasn't picked that change up.

agnivade (approved):

PR #183, opened 2021-04-21
Merge master into production 

hanzei:
#### Summary
In order to sync the two branches I did:
- Create a new branch `merge_production` from `master`
- Merged `production` into that branch
- Fixed merge conflicts
- Run `go run ./cmd/generator/ migrate` to generate the platform specific bundle URLs

Please review for any suspicious looking diffs.

#### Ticket Link
None


hanzei:
This is needed for the Apps plugin release because the `Experimental` label is only on `master`.

catalintomai (approved):

mickmister (approved):

DHaussermann (commented):
Tested this marketplace PR
- Ran through release testing documented in Zephyr for regression testing.
- Explicitly tested mew changes such as Experimental label and improvement to E20/ Cloud label 
- MM Apps functionality
  - No listing shown when not in proxy config
  - No listing shown when a listing in-proxy config does not line up with provisioned Lambda function
  - MM Apps Listing is shown in marketplace as installable
  - MM Apps can be installed via marketplace  
  - MM App can be reinstalled via marketplace if it was installed
  - Apps from manifest are not shown in marketplace when App plugin is disabled 
  
@hanzei Only one point of concern here. Maybe <> has some validity to it?
 Typing into the filter shoes 404 on keystroke when the APPS plugin is disabled. (no issues when it's enabled) but in this case the plugin is installed but disabled. 
 
image

 Is this something that we can address? 
  
Currently Adding MM-Apps test for this to Zephyr as listed above.

DHaussermann:
@amyblais and @aaronrothschild for clarity - This merge will **not** add the Apps plugin to the marketplace. But, it delivers  the needed functionality for the plugin and the lambda applications.

We will need to merge one more PR to get the Apps plugin to the marketplace.

hanzei:
@DHaussermann Please note that the behavior in https://mattermost.atlassian.net/browse/MM-35022 is not related to any changes in this PR.

Still, the behavior is expected. If the feature flag is one (and only if) and the Apps plugin is disabled, the Marketplace fails to load the installed Apps and therefore throws an 404. With the way the feature flag is decoupled from the fact that the Apps plugin is installed enforces this behavior. We can chat about that offline, but it would require substantial changes to the feature flags used for apps to change the behavior.

hanzei:
@DHaussermann Thanks for the testing :tada: I'm going to merge the PR and deffer the discussion on https://mattermost.atlassian.net/browse/MM-35022 to our meeting.

PR #185, opened 2021-04-22
Add v0.4.0 of mattermost-plugin-apps to the Marketplace

hanzei:
#### Summary
Cloud only and experimental 
Screenshot from 2021-04-22 21-29-13

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-apps/issues/153


plugins.json
```diff
@@ -2155,6 +2155,56 @@
...
+    "repo_name": "mattermost-plugin-apps",
+    "manifest": {
+      "id": "com.mattermost.apps",
+      "name": "Apps",
+      "description": "Apps Registry and API proxy.",
+      "homepage_url": "https://github.com/mattermost/mattermost-plugin-apps",
+      "support_url": "https://github.com/mattermost/mattermost-plugin-apps/issues",
+      "release_notes_url": "https://github.com/mattermost/mattermost-plugin-apps/releases/tag/v0.4.0",
+      "version": "0.4.0",
+      "min_server_version": "5.26.0",
```
catalintomai, line 2177: Is this the right version? Or does it matter?

catalintomai (approved):

DHaussermann (commented):
Tested on local deploy 
- Plugin is visible as expected. 
- Plugin shows correct label for experimental   
- Plugin icon and version number are correct
- Plugin does not show for E0, E20 or E20 license 
- Plugin does show for Cloud license
- Plugin can be installed and enabled with cloud licence

1. @hanzei  there is one issue I'm seeing that is a bit odd. Testing Cloud only plugins on a modified URL is awkward as the URL field is read only. I work around this by setting the URL then the Cloud license.

After I install I see the `Local` label and then when I see why, the marketplace is now disabled. 
image

image

Quite likely this is not a bug and the server does not look at that field when you have cloud install? I see nothing like this on production and on the server above the marketplace continues to work. 

2. One more thought you may have some input on. We really need some type of pre-production marketplace. Our Cloud only license check is doing a huge amount of lifting here. This build does not have the min version enforcement yet and there is only one marketplace.

hanzei:
1. Really odd. I suspect that the issue is unrelated to this PR. Would you mind filling a ticket for further investigation?
2. What exactly should be tested on that pre-production marketplace that we can't test on a PR?

aaronrothschild (approved):
Lgtm

DHaussermann:
Created MM-35122 to address this.

hanzei:
@DHaussermann Are you fine with merging this PR or do you want to triage the ticket above first?

DHaussermann:
Okay @hanzei Please merge.

PR #188, opened 2021-05-11
Add v1.10.0 of mattermost-plugin-incident-collaboration to the Market…

lieut-data:
#### Summary
Submit Incident Collaboration 1.10.0 to the marketplace.


prapti (approved):
LGTM!

p.s. Approving before the dev review is done since the changes are straightforward. If there is any change after the dev review however, I can review again.

lieut-data:
Thanks, @prapti! I've added @crspeller as reviewer primarily to satisfy the "2 approving reviews" required. There technically isn't any dev review required here :)

PR #238, opened 2021-11-22
dist: build linux/darwin arm64 targets

cpanato:
#### Summary
Adding the linux/darwin arm64 targets, we start building arm64 for the server and will start adding the targets for the plugins for the ones we prepackage in the build

related to: https://github.com/mattermost/mattermost-server/pull/18491

#### Ticket Link
related to https://mattermost.atlassian.net/browse/DOPS-633


mickmister (commented):
Thanks for making this change @cpanato. You'll also need to run `cmd/generator migrate` to update the existing entries

https://github.com/mattermost/mattermost-marketplace/blob/b03d543bedd82ac728c97892fce20e3fc0193f74/cmd/generator/bundles.go#L27

mickmister:
Actually, disregard the above comment. This will only affect future releases of plugins, which should automatically be updated to the new structure when they are added to the marketplace.

mickmister (approved):

lieut-data (approved):
This seems reasonable, but I'm curious what our plan is to support the new targets. Are we holistically shipping `arm64` and testing it? Do we actually need to bother for Apple's M1 architecture? (I'm assuming no one is running anything but dev servers on it.)

cpanato:
> This seems reasonable, but I'm curious what our plan is to support the new targets. Are we holistically shipping `arm64` and testing it? Do we actually need to bother for Apple's M1 architecture? (I'm assuming no one is running anything but dev servers on it.)

would say we can drop the arm64 for mac but we should keep that for Linux, we will integrate with GitLab omnibus arm674 distribution and in the future use in the cloud as well (to reduce hardware costs)

but we should have a test instance to test that, I will speak with Linda to we sync that, but first lets try to have the packages in place

plugins.json
```diff
@@ -18,7 +18,9 @@
       "server": {
         "executables": {
           "linux-amd64": "server/dist/plugin-linux-amd64",
+          "linux-arm64": "server/dist/plugin-linux-arm64",
```
hanzei, line 21: Do these binaries really exist in 0.1.2? I can't find them in https://github.com/mattermost/mattermost-plugin-antivirus/releases/tag/v0.1.2.

hanzei (commented):
Given that there isn't any cost beside data storage for adding more platform I'm in favor of adding `DarwinArm64`, as that allows dev's to test the marketplace on there own machine.

internal/model/plugin.go
```diff
@@ -130,13 +134,13 @@ func (p *Plugin) AddLabels() {
 
 // PluginFilter describes the parameters used to constrain a set of plugins.
 type PluginFilter struct {
```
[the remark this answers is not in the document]
cpanato: for the lint to organize and be more memory optimized, I can revert

plugins.json
cpanato: no, i was not sure if need to update this file. do I need to revert this?

internal/model/plugin.go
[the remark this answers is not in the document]
hanzei: Please revert. The original ordering was easier to understand.

plugins.json
hanzei: Please do. I don't think this change is correct.

cpanato: ahh, looking at it again, that is completely wrong. removed my fault

hanzei (approved):

hanzei:
@DHaussermann gentle reminder to review this PR. Please let me or Carlos how, if you need a testing device or test person.

DHaussermann:
@cpanato or @hanzei I see that unit tests have been added. For functional testing on this, it seems I would need a Mattermost servers running on hardware with ARM architecture for Linux and OS X on to confirm that only the correct binary is sent from the marketplace.  

I don't know of any such test servers. Do you know of any infrastructure like this that is available?

cpanato:
there is no infra yet with arm64 machines, but you can spin one and try out

hanzei:
Adding support for linux/darwin arm64 targets is not on the priority list currently. We can resurface this PR if needed.

PR #249, opened 2022-02-17
Add v0.14.0 of focalboard to the Marketplace

sbishel:
#### Summary
Publish Focalboard (Boards) v0.14.0 to the Marketplace


chenilim (approved):
I'm pretty sure this is right, but just want to double-check that we didn't update the min-server requirement in 0.14 right? Otherwise, LGTM!

hanzei (approved):

PR #268, opened 2022-04-06
[MM-42841] Correctly set ReturnAllVersions and PluginID for upstream requests

hanzei:
#### Summary
In https://github.com/mattermost/mattermost-marketplace/pull/142  it was forgotten to add  `ReturnAllVersions` and 
`PluginID` to `GetPluginsRequest`. Hence, the request to upstream didn't contain these fields and the upstream response would contain all plugins, not only the requested one.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42841


.golangci.yml
```diff
@@ -18,6 +18,8 @@ linters-settings:
   govet:
     check-shadowing: true
     enable-all: true
+    disable:
+      - fieldalignment
```
hanzei, line 22: We are already ignoring it in other projects, let's also ignore it here

hanzei:
cc @streamer45

mickmister (approved):

catalintomai (approved):

streamer45:
@hanzei For my own understanding, could you explain to me how this fix is related to the issue we were facing earlier in https://community-daily.mattermost.com/core/pl/y996dt987inrudqw9se51srmxe? The problem we had with Calls was that the plugin didn't exist on the Marketplace but was pre-packaged so we expected it to be installable anyway but since a subsequent check was added in https://github.com/mattermost/mattermost-server/pull/19613 the install request would fail.

hanzei:
@streamer45 Maybe I've mistaken https://mattermost.atlassian.net/browse/MM-42841 as the root cause for your issue. I will look further into this and let you know.

hanzei:
@streamer45 Comming back late to this: The cause of the issue you are seeing is https://mattermost.atlassian.net/browse/MM-45296. There will be a seperate PR to fix that.

hanzei:
Given that this is a straight forward change which is covered by a test I'm going to merge it without QA approval.

cc @DHaussermann

DHaussermann:
Got it 👍 
Thanks @hanzei

PR #272, opened 2022-04-19
MM-43497 - Add on-prem label to marketplace calls plugin

cpoile:
#### Summary
- Calls is only available for on-prem installations at the moment.
- I think this should be all we need, @hanzei can you confirm?

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-43497


streamer45 (approved):
Thanks for prioritizing this :+1:

hanzei (approved):
That does the trick :+1:

PR #277, opened 2022-05-04
Add v0.4.9 of mattermost-plugin-calls to the Production Marketplace

cpoile:
#### Summary
- cut with `/mb cutplugin --repo mattermost-plugin-calls --tag v0.4.9 --pre-release`
- this commit made with `go run ./cmd/generator/ add mattermost-plugin-calls v0.4.9 --official --beta --on-prem`

#### Ticket Link
- none

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -3303,6 +3303,101 @@
   {
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
     "icon_data": "",
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-calls-v0.4.9.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-calls/releases/tag/v0.4.9",
+    "hosting": "on-prem",
+    "author_type": "mattermost",
+    "release_stage": "beta",
+    "enterprise": false,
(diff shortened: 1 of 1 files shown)


plugins.json
```diff
@@ -3300,6 +3300,101 @@
...
+            "default": 8443
+          },
+          {
+            "key": "RTCDServiceURL",
+            "display_name": "RTCD Service URL",
+            "type": "text",
+            "help_text": "The URL to a running RTCD service instance that should host the calls. When set (non empty) all calls will be handled by the external service.",
+            "placeholder": "https://rtcd.example.com",
+            "default": null
+          },
```
streamer45, line 3360: This makes me think a little as we are most likely not going to officially support `rtcd` on prem for the beta launch. But I guess there isn't a clean way to hide certain settings when not in Cloud?

streamer45 (approved):
Thanks 👍

plugins.json
cpoile: @streamer45 I don't think we have that ability -- we could add some explanation there if you're worried about it?

streamer45: Yeah, no worries. I think it's okay for the beta as supporting `rtcd` in on-prem is more about properly documenting everything at this point. Implementation wise it would work already so it's good to have it in sooner rather than later.

cpoile:
Needs two reviews, so tagging @lieut-data. :)

lieut-data (approved):

PR #281, opened 2022-05-24
Add v1.1.0 of mattermost-plugin-apps to the Marketplace

levb:
Release notes at https://github.com/mattermost/mattermost-plugin-apps/releases/tag/v1.1.0

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -2763,7 +2763,7 @@
     "homepage_url": "https://mattermost.com/marketplace/app-framework",
     "icon_data": "",
     "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-apps-v1.1.0.tar.gz",
-    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-apps/releases/tag/v1.0.1",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-apps/releases/tag/v1.1.0",
     "hosting": "",
     "author_type": "mattermost",
     "release_stage": "production",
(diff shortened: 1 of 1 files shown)


plugins.json
```diff
@@ -2759,6 +2759,56 @@
     },
     "updated_at": "2020-11-20T18:12:44.537612606Z"
   },
+  {
+    "homepage_url": "https://mattermost.com/marketplace/app-framework",
+    "icon_data": "",
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-apps-v1.1.0.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-apps/releases/tag/v1.0.1",
```
mickmister, line 2766: Looks like this was missed in https://github.com/mattermost/mattermost-plugin-apps/pull/345. Maybe something we can automate?

```suggestion
    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-apps/releases/tag/v1.1.0",
```

mickmister (changes requested):

plugins.json
levb: Right, was overlooked in https://github.com/mattermost/mattermost-plugin-apps/pull/345.

I wonder if we can just autogenerate `release_notes_url` and leave it empty in `plugin.json`

Can we leave with the incorrect URL in the plugin itself and not re-cut the release? I can fix this PR so that we have the right URL in the marketplace.

hanzei (approved):

plugins.json
mickmister: > Can we leave with the incorrect URL in the plugin itself and not re-cut the release? I can fix this PR so that we have the right URL in the marketplace.

@levb Yep works for me

mickmister: > I wonder if we can just autogenerate `release_notes_url` and leave it empty in `plugin.json`

This would be a "update all plugin repos" change, which has proven to not scale well. Having a common tooling source may be a better way to go about managing the plugin repository tooling. @hanzei What do you think about this?

catalintomai (approved):

mickmister:
Should this be merged/closed?

hanzei:
We don't need this as the plugin is pre packaged

PR #288, opened 2022-06-07
Add v0.6.0 of mattermost-plugin-calls to the Marketplace

cpoile:
#### Summary
- cut with `/mb cutplugin --repo mattermost-plugin-calls --tag v0.6.0--pre-release`
- this commit made with `go run ./cmd/generator/ add mattermost-plugin-calls v0.6.0 --official --beta --on-prem`

#### Ticket Link
- none

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -3303,6 +3303,109 @@
   {
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
     "icon_data": "",
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-calls-v0.6.0.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-calls/releases/tag/v0.6.0",
+    "hosting": "on-prem",
+    "author_type": "mattermost",
+    "release_stage": "beta",
+    "enterprise": false,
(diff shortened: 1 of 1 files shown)


plugins.json
```diff
@@ -10379,23 +10482,23 @@
...
-    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-zoom-v1.5.1.tar.gz",
-    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-zoom/releases/tag/v1.5.1",
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-zoom-v1.6.0.tar.gz",
```
lieut-data, line 10485: Why is the Zoom plugin changing?

lieut-data (changes requested):

plugins.json
```diff
@@ -10499,40 +10602,40 @@
...
-    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-zoom-v1.6.0.tar.gz",
-    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-zoom/releases/tag/v1.6.0",
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-zoom-v1.5.1.tar.gz",
```
streamer45, line 10622: Is this expected?

plugins.json
cpoile: @lieut-data I have no idea how that got in there. It wasn't in my local diff...

cpoile:
The generator is broken for this one, it keeps updating zoom as well. I haven't taken the time to look at the code, though. If it happens again I will.

streamer45:
@cpoile We could also hold off this PR if you are worried. I don't think there's any hard deadline for this other than June 15th.

lieut-data:
> The generator is broken for this one, it keeps updating zoom as well. I haven't taken the time to look at the code, though. If it happens again I will. I kind of want to figure out what's going on, but I'm afraid it will take up too much time...

I think it would be fine to just revert this portion of the changes and use only the part re: calls. We aren't required to pull in all the new marketplace changes.

cpoile:
@streamer45 I think we're good to go, not too worried about it. Just wondering what happened. :)

plugins.json
```diff
@@ -3300,6 +3300,109 @@
...
+            "placeholder": "stun:example.com:3478",
+            "default": "stun:stun.global.calls.mattermost.com:3478"
+          },
+          {
+            "key": "RTCDServiceURL",
+            "display_name": "RTCD service URL",
+            "type": "text",
+            "help_text": "(Optional) The URL to a running RTCD service instance that should host the calls. When set (non empty) all calls will be handled by the external service.",
+            "placeholder": "https://rtcd.example.com",
+            "default": null
```
streamer45, line 3391: Hope this won't cause issues, it shouldn't really.

lieut-data (approved):

plugins.json
cpoile: Ah, it's because we don't have a `default` specified. Up to you, we could manually change it here then change it in code later.

streamer45 (approved):

PR #292, opened 2022-06-10
Add v0.6.1-alpha2 of mattermost-plugin-calls to the Marketplace

cpoile:
#### Summary
- This is for testing the cloud free trial changes in https://github.com/mattermost/mattermost-plugin-calls/pull/105 That's why it's `"hosting": "cloud"`
- Will be reverted as soon as testing is done

#### Ticket Link
- none

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -3303,6 +3303,109 @@
   {
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
     "icon_data": "",
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-calls-v0.6.1-alpha2.tar.gz",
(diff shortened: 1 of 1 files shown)


lieut-data (approved):
I wonder if we need a "cloud-test" marketplace overlay like private-beta was? As it stands, this can be installed by customers with access to the marketplace, right?

cpoile:
@lieut-data I believe it will not show up for on-prem customers, because of the `cloud` flag, correct? But yes, cloud customers will be able to install it. (no big deal if they do, I think) 
Yes, we do need a private marketplace, and we need to make sure preinstalled plugins can be pulled from there.

streamer45 (approved):

PR #326, opened 2022-12-14
[MM-49087] Add v1.3.1 of mattermost-plugin-nps to the Marketplace

JulienTant:
#### Summary
Add v1.3.1 of mattermost-plugin-nps to the Marketplace
https://github.com/mattermost/mattermost-plugin-nps/pull/99

#### Ticket Link
[mattermost.atlassian.net/browse/MM-49087](https://mattermost.atlassian.net/browse/MM-49087)


hanzei (approved):

hanzei:
Does the new release also need to be prepackaged?

hanzei:
Just found https://github.com/mattermost/mattermost-server/pull/21885

mickmister (approved):

marianunez (commented):
@esethna Is there any value of having this in the marketplace? This is a pre-packaged plugin for our feedback purposes, so not sure if it makes sense to add to the marketplace as a choice for admins

esethna:
+1, we don't need to add this in the marketplace assuming it's prepackaged, turned on by default, and updated automatically. @marianunez @JulienTant

JulienTant:
Then should we remove it from the marketplace altogether? Because it's present right now but outdated.

esethna:
I would say yes, as long as admins can disable/remove it from the sys console

marianunez:
> I would say yes, as long as admins can disable/remove it from the sys console

@esethna They are able to disable/remove it but wouldn't be able to bring it back from the Marketplace (which is very unlikely they would want to do so). Wonder if there is any data on downloads for NPS from Marketplace

PR #332, opened 2023-01-09
MM-49042 - Add v0.12.0 of mattermost-plugin-calls to the Marketplace

cpoile:
#### Summary
- NOTE: I had to do this kind of manually -- the marketplace code in the production branch is very out of date (it's using a manifest model from mattermost v5). So I went to the master branch, updated the model to a recent v6 from December, which has the changes we need (the "hosting" field), ran the code from there, and manually cut and pasted the json to a branch based on production. 
- I'll submit a pr to the marketplace in master, then we'll decide what to do to bring production up to speed. 

- cut with `/mb cutplugin --repo mattermost-plugin-calls --tag v0.12.0`
- this commit made with `go run ./cmd/generator/ add mattermost-plugin-calls v0.12.0 --official`

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49042

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -3676,7 +3676,7 @@
           "support_url": "https://github.com/mattermost/mattermost-plugin-calls/issues",
           "release_notes_url": "https://github.com/mattermost/mattermost-plugin-calls/releases/tag/v0.12.0",
           "version": "0.12.0",
-          "min_server_version": "7.6.0",
+          "min_server_version": "7.7.0",
           "server": {
               "executables": {
                   "darwin-amd64": "server/dist/plugin-darwin-amd64",
--- a/plugins.json
+++ b/plugins.json
@@ -3658,6 +3658,177 @@
     "updated_at": "2019-06-11T19:41:07Z"
   },
   {
+      "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
+      "icon_data": "",
+      "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-calls-v0.12.0.tar.gz",
+      "release_notes_url": "https://github.com/mattermost/mattermost-plugin-calls/releases/tag/v0.12.0",
+      "hosting": "",
+      "author_type": "mattermost",
+      "release_stage": "production",
+      "enterprise": false,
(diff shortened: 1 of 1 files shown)


plugins.json
```diff
@@ -3657,6 +3657,177 @@
...
+          "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
+          "support_url": "https://github.com/mattermost/mattermost-plugin-calls/issues",
+          "release_notes_url": "https://github.com/mattermost/mattermost-plugin-calls/releases/tag/v0.12.0",
+          "version": "0.12.0",
+          "min_server_version": "7.6.0",
+          "server": {
+              "executables": {
+                  "darwin-amd64": "server/dist/plugin-darwin-amd64",
+                  "darwin-arm64": "server/dist/plugin-darwin-arm64",
+                  "freebsd-amd64": "server/dist/plugin-freebsd-amd64",
```
lieut-data, line 3684: Huh, when did freebsd and openbsd become a thing?

plugins.json
```diff
@@ -3657,6 +3657,177 @@
...
+      "repo_name": "mattermost-plugin-calls",
+      "manifest": {
+          "id": "com.mattermost.calls",
+          "name": "Calls",
+          "description": "Integrates real-time voice communication in Mattermost",
+          "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
+          "support_url": "https://github.com/mattermost/mattermost-plugin-calls/issues",
+          "release_notes_url": "https://github.com/mattermost/mattermost-plugin-calls/releases/tag/v0.12.0",
+          "version": "0.12.0",
+          "min_server_version": "7.6.0",
```
lieut-data, line 3679: nit: given 7.6 doesn't really exist, should this be 7.7?

lieut-data (approved):

plugins.json
streamer45: We have community instances on both of these platforms running Calls as we speak:

https://github.com/mattermost/mattermost-plugin-calls/issues/45
https://github.com/mattermost/mattermost-plugin-calls/issues/97

plugins.json
streamer45: That's a good point. I am not opposed to change it here but to have it in the plugin we'd have to go back and restart the release process. Not sure if worth it but happy to hear thoughts. @cpoile?

cpoile: I don't mind the extra work if it makes it cleaner. But on the other hand, nothing will break, so 🤷 . Up to you @streamer45

streamer45: Let's change it here then and will fix it on the plugin side starting in next release.

streamer45 (approved):

cpoile:
I don't know why it dismissed Jesse's review as "stale" -- maybe because I pushed a new commit? New rules?  Sorry @lieut-data

streamer45:
> I don't know why it dismissed Jesse's review as "stale" -- maybe because I pushed a new commit? New rules? Sorry @lieut-data

Yeah, sounds like the same repo setting as we have in [`rtcd`.](https://github.com/mattermost/rtcd)

plugins.json
lieut-data: Wow, cool!

lieut-data (approved):

PR #333, opened 2023-01-11
MM-49609 - Update mattermost-server version

cpoile:
#### Summary
- Calls is using the `hosting` field in the `setting_schema.settings` objects, so we need a more recent version of the server’s `model` in order to parse that field and include it in the `plugins.json`.
- Also requires updating production branch, which I'll do in a separate pr.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49609


streamer45:
@cpoile Just so you know build is failing.

streamer45:
> @streamer45 Yah, I can't see why -- I fixed make check-style and now it runs fine locally. Have you seen this before, @hanzei ?

@cpoile I think we should bump Golang version of the runner.

streamer45 (approved):
Nice :+1:

.circleci/config.yml
```diff
@@ -39,8 +39,6 @@ jobs:
     working_directory: *workspace_root
     steps:
       - checkout
-      - tools/install-golangci-lint:
-          version: 1.33.2
```
hanzei, line 43: Do we still want to control the golangci-lint version? I'm wondering if it's valuable to keep the version in sync with the one used in plugins. https://github.com/mattermost/circleci-orbs/blob/ea9f34540218ce3a62b101eab2c4fffc331be11b/plugin-ci/orb.yml#L51-L53

.circleci/config.yml
```diff
@@ -1,15 +1,15 @@
...
 
 orbs:
-  tools: gotest/tools@0.0.13
+  tools: gotest/tools@0.0.14
 
 executors:
   build:
     docker:
-      - image: circleci/golang:1.16
+      - image: cimg/go:1.18.8
```
hanzei, line 9: While we are here: Should the version be bumped to 1.19?

.golangci.yml
```diff
@@ -39,24 +36,20 @@ linters:
     - gocyclo
     - gofmt
     - goimports
-    - golint
```
hanzei, line 42: Other projects use `revive` as a replacement for `golint`. Should we do the same in this project?

hanzei (changes requested):
Thanks for keeping the dependencies up to date :+1: 

A few questions below

.circleci/config.yml
cpoile: sure, done

cpoile:
@hanzei gentle ping on this, thank you!

cpoile:
@hanzei can you take another look at this at your earliest convenience? We want to merge this and then add another PR to merge master->production so that our release process is unblocked. Thanks!

.circleci/config.yml
hanzei: 3/5 it's, in general, helpful to use the same `golangci-lint` version across multiple projects as there are fewer "surprises" when CI runs. A dev can only install one version locally, and that should ideally match one used in CI. For (most) plugins, the centralised CI config already synchronises the version used.

On the other hand, it's cumbersome to update the version for all other projects at the same time. Hence, I suggested we use this opportunity to sync the version with the one used in the plugin repos.

hanzei (approved):
One optional suggestion, other LGTM :+1:

hanzei:
@cpoile Sorry for the delay.

.circleci/config.yml
cpoile: @hanzei Thanks! 

Ok, I figured how to get the up to date `golangci-lint` -- it's included in the `cimg/go` container already, so no need for the `gotest/tools`. Probably the `plugin-ci` orb can be simplified (we don't need to install `golangci-lint` there either)

cpoile: I'll merge now so I can start with the production branch merge.

PR #334, opened 2023-01-11
MM-49610 - Merge master into production

cpoile:
#### Summary
- Related to #333 , merge master into production because we (calls) run the generator cmd on the production branch
- The second commit is updating the existing plugins to the new manifest (an empty `hosting` field). This way the next person to run the generator won't see all this in their diff.
- I don't _think_ this will cause any troubles with the marketplace, but correct me if I'm wrong please.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49610


cpoile:
Sorry -- need to double check the commits, it should be cleaner than this

cpoile:
I'm going to wait until #334 is merged, then do this. Apologies for the noise.

Issue #118, opened 2020-09-17
Serverless version for pipelines doesn't fulfill  serverless config


hanzei:
Serverless v2.0.0 was released 2 days ago. Because the circleci pipeline does install the latest version it installs `>= v2.0.0` which is incompatible with `serverless.yml`.

This causes all deployments to fail: https://app.circleci.com/pipelines/github/mattermost/mattermost-marketplace/425/workflows/80fb9c71-3b6c-4af7-b0e5-87fa7503a7ff/jobs/536

PR #338, opened 2023-01-31
MM-50203 - Update serverless

cpoile:
#### Summary
- Update serverless, following the advice of: https://github.com/serverless/serverless/issues/11118

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-50203

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,3 +79,8 @@ workflows:
           context: mattermost-marketplace
           requires:
             - build
+          filters:
+            branches:
+              only:
+                - master
+                - production
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,8 +79,3 @@ workflows:
           context: mattermost-marketplace
           requires:
             - build
-          filters:
-            branches:
-              only:
-                - master
-                - production
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
       - *attach_workspace
       - run:
           name: Install Serverless CLI and dependencies
-          command: sudo npm i -g serverless@"<2.0.0"
+          command: sudo npm i -g serverless@3
       - run:
           name: Deploy Function
           command: |


.circleci/config.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - *attach_workspace
       - run:
           name: Install Serverless CLI and dependencies
-          command: sudo npm i -g serverless@"<2.0.0"
+          command: sudo npm i -g serverless@3
```
hanzei, line 56: Did you test that the pipelien works with the change? Serverless v3 has caused issues in the past: https://github.com/mattermost/mattermost-marketplace/issues/118

hanzei (changes requested):
Thanks for jumping on the issue :+1:

.circleci/config.yml
cpoile: @hanzei Thanks. I'm not sure how to run the pipeline without committing and trying. Please let me know if you know of another way.

hanzei: If you go to a the pipeline run on `master` https://app.circleci.com/pipelines/github/mattermost/mattermost-marketplace/979/workflows/2fdc5893-9ad2-4a0f-bd87-4590302ba0b9/jobs/1273, you can run a job with SSH and then SSH onto the machine
Screenshot from 2023-02-02 10-07-19

That allows you to test your changes.

hanzei: Please let me know if you nee another guidance.

cpoile: Yep, I know about rerunning workflows. The workflow that is failing is the `deploy`, which this PR doesn't have. So rerunning wouldn't test it. I guess I'll add deploy to this branch to test it, and then remove.

cpoile: @hanzei Ok. SSHing won't help if the deploy workspace isn't there, so I added the deploy branch to this PR, and tested twice -- serverless installs without a problem. However, I can't verify the rest of the deploy script works in this branch. I suggest we merge to master and see if it completes there.

hanzei (approved):
Sure, CD can't be more broken then it as right now

PR #337, opened 2023-01-27
MM-49610 - Merge master into production

cpoile:
#### Summary
- Related to #333 , merge master into production because we (calls) run the generator cmd on the production branch
- After all the commits from Master, there are two new ones: the merge commit [b248b8f](https://github.com/mattermost/mattermost-marketplace/pull/337/commits/b248b8fbb9e0af245b3dba1828bc13a849012ec9), and the `plugins.json` diff [ea51c3e](https://github.com/mattermost/mattermost-marketplace/pull/337/commits/ea51c3eb1a0792f38ce0b68ddbc661632ddd693a)
- [ea51c3e](https://github.com/mattermost/mattermost-marketplace/pull/337/commits/ea51c3eb1a0792f38ce0b68ddbc661632ddd693a) is updating the existing plugins to the new manifest (an empty `hosting` field). This way the next person to run the generator won't see all this in their diff.
- I don't _think_ this will cause any troubles with the marketplace, but correct me if I'm wrong please.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49610

PR #347, opened 2023-03-09
MM-50566 - Add mattermost plugin calls v0.14.0

cpoile:
#### Summary
- NOTE: I had to do this kind of manually, again (see #337)
- cut with `/mb cutplugin --repo mattermost-plugin-calls --tag v0.14.0`
- this commit made with `go run ./cmd/generator/ add mattermost-plugin-calls v0.14.0 --official`

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-50566

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -4088,6 +4088,187 @@
   {
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
     "icon_data": "",
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-calls-v0.14.0.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-calls/releases/tag/v0.14.0",
+    "hosting": "",
+    "author_type": "mattermost",
+    "release_stage": "production",
+    "enterprise": false,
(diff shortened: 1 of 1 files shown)


streamer45 (approved):

plugins.json
```diff
@@ -4085,6 +4085,187 @@
...
+      "min_server_version": "7.6.0",
+      "server": {
+        "executables": {
+          "darwin-amd64": "server/dist/plugin-darwin-amd64",
+          "darwin-arm64": "server/dist/plugin-darwin-arm64",
+          "freebsd-amd64": "server/dist/plugin-freebsd-amd64",
+          "linux-amd64": "server/dist/plugin-linux-amd64",
+          "linux-arm64": "server/dist/plugin-linux-arm64",
+          "openbsd-amd64": "server/dist/plugin-openbsd-amd64"
+        },
```
mickmister, line 4116: At the same time, it seems these these are all compiled, so I would expect them to be in the individually-built binaries referenced in the above comment. I see something related to calls in matterbuild's scripts. I'm not sure if this is still occurring in the most up-to-date matterbuild though

https://github.com/mattermost/matterbuild/blob/59f5f2df3571c981871ed21cf78a420d1934e539/server/plugin_release.go#L411-L416

mickmister (approved):

plugins.json
mickmister: Looks like the "individually building" script needs to be updated in general, to support more platforms. The corner case for Calls is right next to that code.

mickmister:
@cpoile Feel free to make the decision on QA review or not :+1:

cpoile:
We haven't in the past, so given the reduced QA cap, I'm thinking we're fine merging.

plugins.json
cpoile: Not sure what you mean? The matterbuild code builds and verifies the package, and that's what the generator parsed and output into the `plugins.json`.

mickmister: In the comment above https://github.com/mattermost/mattermost-marketplace/pull/347#discussion_r1131493026, the `darwin-amd64` and `windows-amd64`-specific bundles are missing in the new marketplace entry. This is because matterbuild is not individually packaging those "single-architecture" bundles.

I'm saying that:
- The general case of packaging these this way should take account making single-architecture bundles for all 6 architectures we now support, instead of just the original 3 we supported
- and the corner case for calls also should do something similar, since it is only packaging one single-architecture bundle at the moment

Just so I understand clearly, the calls plugin indeed runs on the above 6 architectures? I think at some point this wasn't the case so just checking.

The single-architecture bundle is an optimization we made for spinning up Cloud instances quicker (much smaller bundle to download), so I don't think the above points are required changes. The multi-architecture bundles are bigger, but the benefits are probably negligible for on-prem deployments.

streamer45: @mickmister Calls is only officially supported on Linux (amd64). Any other target was either added by community contributors (open/freebsd) and darwin is used for development only so it shouldn't get shipped in any official build.

PR #363, opened 2023-04-13
Add v3.2.0 of mattermost-plugin-azure-devops to the Marketplace

mickmister:


plugins.json
```diff
@@ -15723,6 +15723,99 @@
...
+      "icon_path": "public/assets/azurebot.svg",
+      "version": "3.2.0",
+      "min_server_version": "5.37.3",
+      "server": {
+        "executables": {
+          "darwin-amd64": "server/dist/plugin-darwin-amd64",
+          "darwin-arm64": "server/dist/plugin-darwin-arm64",
+          "linux-amd64": "server/dist/plugin-linux-amd64",
+          "linux-arm64": "server/dist/plugin-linux-arm64",
+          "windows-amd64": "server/dist/plugin-windows-amd64.exe"
```
trilopin, line 15753: not familiar with this but, is it expected to have the 5 binaries here but only 3 in the platforms section?

plugins.json
```diff
@@ -15723,6 +15723,99 @@
...
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-azure-devops-v3.2.0.tar.gz",
```
trilopin, line 15729: out of curiosity (comparing this to autolink), why do we have much more data for azure than autolink (signatures, platform details, icon data, etc..)

plugins.json
mickmister: @trilopin These 5 executables are what the plugin was compiled to support. In the main bundle of the plugin's release, the `tar.gz` file contains all server executables listed here.

The `platforms` struct you're mentioning is an optimization that is currently out of sync with all the architectures we support. There's an open PR to synchronize these here https://github.com/mattermost/mattermost-marketplace/pull/238. During the execution of the `/mb cutplugin` command, Matterbuild [extracts out each executable individually](https://github.com/mattermost/matterbuild/blob/59f5f2df3571c981871ed21cf78a420d1934e539/server/plugin_release.go#L404-L408), and uploads these more bare bundles to S3. The URLs to the CDN serving those bundles are then used in the `platforms` struct.

This was originally done to improve Cloud server startup times, since syncing with the file store on startup was causing a lot of delay on startup. Since this feature was not made to help self-hosted servers, it hasn't been diligently maintained as such. Older servers that don't know about the optimization, and servers that are running on other architectures, are unaffected since it is optional and backwards-compatible.

trilopin (approved):

PR #395, opened 2023-11-15
Add v2.1.7 of mattermost-plugin-github to the Marketplace.

avas27JTG:
#### Ticket Link
- https://github.com/mattermost/mattermost-plugin-github/pull/708


plugins.json
```diff
@@ -7111,6 +7111,194 @@
...
+    "download_url": "https://plugins-store.test.mattermost.com/release/mattermost-plugin-github-v2.1.7.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-github/releases/tag/v2.1.6",
```
hanzei, line 7118: That is the wrong URL. Do we need to fix the release?

avas27JTG: @hanzei @mickmister I've created this PR https://github.com/mattermost/mattermost-plugin-github/pull/716 and after merging it I can cherry-pick this change into the tag `v2.1.7` to fix this.
Please let me know if it sounds good or we need to re-create this release with this change.

mickmister: We should probably have a check for this in the `marketplace generate` command

mickmister: @avas27JTG I think we can just edit the release note in-place in the marketplace entry here. No need to go through the release process of the plugin again. I don't think the marketplace will complain about this inconsistency

mickmister, line 7118:
```suggestion
    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-github/releases/tag/v2.1.7",
```

mickmister: I tried to edit the file through review suggestion, but it won't let me because of the PR permissions thing

avas27JTG: Fixed it.

avas27JTG:
> Blocking on [#395 (comment)](https://github.com/mattermost/mattermost-marketplace/pull/395#discussion_r1409761763)

@hanzei its fixed now

PR #398, opened 2023-11-17
MM-55596 - Remove Calls v0.21.0 from the marketplace

cpoile:
#### Summary
- v0.21.0 has a bug (in that it is sending push proxy message types that are not supported). Customers should use v0.21.1 instead.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-55596


streamer45:
@cpoile Is there any instance in which a user would be able to install the previous release from the Marketplace UI? I assumed we would always return the latest available version that's supported by the MM instance.

cpoile:
@streamer45 I do not know.. I just want to be thorough and remove it (along with the artifacts and the git tag).

cpoile:
Discussed with Claudio offline, giong to close this -- customers know to upgrade dot releases, and our v0.21.1 fix was published a few days after the first.

lieut-data:
Fine to ignore this, but to answer:

> Is there any instance in which a user would be able to install the previous release from the Marketplace UI? I assumed we would always return the latest available version that's supported by the MM instance.

Strictly speaking, we can use the MM API to install a specific version from the marketplace. So this is still "installable". But so is downloading the old version from GitHub.

PR #402, opened 2023-12-05
Update private beta

lieut-data:
#### Summary
Update the `private-beta` branch from `master`, resetting `plugins.json` to the nil set as well as updating the deployment code to use GitHub Actions.

#### Ticket Link
None.


lieut-data:
@mvitale1989 & @mickmister, most of this should just bring the branch up-to-date with master, though feel free to compare for confirmation: https://github.com/mattermost/mattermost-marketplace/compare/master...update-private-beta (files changed).

I'll comment inline for where the differences are.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.MM_MARKETPLACE_AWS_SECRET_ACCESS_KEY_STAGING }}
+      run: serverless deploy function -f server --stage staging --verbose
+    - name: deploy/production-branch
+      if: ${{ github.ref_name == 'production' }}
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.MM_MARKETPLACE_AWS_ACCESS_KEY_ID_PRODUCTION }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.MM_MARKETPLACE_AWS_SECRET_ACCESS_KEY_PRODUCTION }}
+      run: serverless deploy function -f server --stage production --verbose
+    - name: deploy/private-beta-branch
+      if: ${{ github.ref_name == 'private-beta' }}
```
lieut-data, line 78: Added new deploy step, only in this branch.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+        }
+    - name: build/persist-build
+      uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+      with:
+        name: marketplace-build
+        path: dist/
+        retention-days: 2
+
+  deploy:
+    if: ${{ github.repository == 'mattermost/mattermost-marketplace' && (github.ref_name == 'master' || github.ref_name == 'production' || github.ref_name == 'private-beta') }}
```
lieut-data, line 45: Added `private-beta` ref check.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+      with:
+        fetch-depth: 0
+    - name: build/ensure-master-exists
+      run: git rev-parse --verify master >/dev/null 2>&1 || git branch master origin/master
+    - name: build/check-style
+      run: make check-style
+    - name: build/test
+      run: make test
+    - name: build/build
+      if: ${{ github.repository == 'mattermost/mattermost-marketplace' && (github.ref_name != 'private-beta') }}
```
lieut-data, line 24: Added `if` check for `!= private-beta`.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+      run: make build
+    - name: build/build-private-beta
```
lieut-data, line 26: Added build target for `private-beta` that essentially just sets `BUILD_UPSTREAM_URL`.

plugins.json
lieut-data, line 1: File empty in this branch, for now. Just means it should exactly match upstream production.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+      if: ${{ github.repository == 'mattermost/mattermost-marketplace' && (github.ref_name == 'private-beta') }}
+      run: BUILD_UPSTREAM_URL=https://api.integrations.mattermost.com make build
```
mickmister, line 28: It took me a second to realize what's happening between these two steps. Both steps run `make build`, but with slightly different contexts. The first runs for any branch that has these changes, besides the `private-beta` branch itself, and the second runs just for the `private-beta` branch, but with an added env var for the upstream endpoint. What's the main benefit of /not/ having the `BUILD_UPSTREAM_URL` set for the first step?

Also the second step may be a bit easier to read if we put this `BUILD_UPSTREAM_URL` in a `env` block of this step, rather than putting it in the `run` statement before `make build`. The `make build` is a bit de-emphasized at the moment.

mickmister (approved):
Awesome work @lieut-data :+1:

LGTM, just one comment for discussion

.github/workflows/build-and-deploy.yml
lieut-data: For the first build, it's imperative that `BUILD_UPSTREAM_URL` isn't set, since we don't want the production server making a request back to itself (or any other marketplace instance). In principle, we could make the bit to check an upstream instance not require a custom build and instead do this at run time -- back in the day, this was easiest to me :p

Your second suggestion is much appreciated! I've updated the CI to use `env: ` to fix this -- thank you :)

mvitale1989 (approved):
I reviewed [the diff that you posted](https://github.com/mattermost/mattermost-marketplace/compare/master...update-private-beta) and it looks good to me, thanks Jesse! :rocket: 

A nonblocking question on this: there seem to be very few differences between `master` and `private-beta`, would it make sense to "unify" those branches somehow? E.g. parameterizing the plugin list during the build, to be able to build both from the same commit. Then we could periodically move `private-beta` to the `master` branch tip (kinda like the cloud branch used to do in the monorepo).
This would have the benefit of unifying the GHA workflow as well, while right now we have diverging versions for each branch.
Ideally it'd be nice to do the same for `production` too, but I see there are many more differences compared to `master`.

lieut-data:
@mvitale1989, I think what you describe is perfectly reasonable. That being said, the architecture of this whole project was extremely "light", down to the database just being a `.json` file -- it hasn't materially changed in the past 4 years, and even though the branches get out of sync, I suspect the overhead of the occasional merge like this is still relatively low. I'd be inclined to keep it simple until such time as it becomes a pain point.

PR #421, opened 2024-03-18
Add v3.2.1 of mattermost plugin azure devops to the Marketplace

raghavaggarwal2308:
#### Summary
- https://github.com/mattermost/mattermost-plugin-azure-devops/pull/161
- Tested by exposing local marketplace with ngrok and adding to a cloud server.


hanzei (commented):
Azure DevOps was removed from the Marketplace a few months ago. Do we really want to add it again?

raghavaggarwal2308:
@hanzei Apologies for that, I did not remember that. We can close the PR if that is the case.

PR #440, opened 2024-05-17
Fix GH action workflow for notifying incoming webhooks on community/hub servers

mickmister:
#### Summary

This PR uses newly set environment variables with incoming webhooks to send release updates to the community/hub servers.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/CLD-7673

https://hub.mattermost.com/private-core/pl/cdque38knirqpet1soc8x1398e

PR #441, opened 2024-05-17
Production: Fix GH action workflow for notifying incoming webhooks on community/hub servers

mickmister:
#### Summary

This PR uses newly set environment variables with incoming webhooks to send release updates to the community/hub servers.

This PR is for the production branch. Here's the one for the master branch https://github.com/mattermost/mattermost-marketplace/pull/440

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/CLD-7673

https://hub.mattermost.com/private-core/pl/cdque38knirqpet1soc8x1398e

### DIFF
--- a/.github/workflows/notify-mm-blog.yml
+++ b/.github/workflows/notify-mm-blog.yml
@@ -17,12 +17,11 @@ jobs:
         shell: bash
         run: |
           .github/workflows/notify-mm-blog.sh
-      - name: Notify channel 1
+      - name: Notify "Community Configuration" channel on community
         uses: mattermost/action-mattermost-notify@master
         env:
-          MATTERMOST_WEBHOOK_URL: ${{ secrets.MATTERMOST_WEBHOOK_URL_1 }}
-      - name: Notify channel 2
+          MATTERMOST_WEBHOOK_URL: ${{ secrets.MM_COMMUNITY_COMMUNITY_CONFIGURATION_INCOMING_WEBHOOK }}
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-mm-blog.yml
```diff
@@ -17,12 +17,11 @@ jobs:
         shell: bash
         run: |
           .github/workflows/notify-mm-blog.sh
-      - name: Notify channel 1
+      - name: Notify "Community Configuration" channel on community
         uses: mattermost/action-mattermost-notify@master
         env:
```
saturninoabril, line 22: This should be `with` instead of `env`. Also might need to define `PAYLOAD`. See new usage for v2.0/master - https://github.com/mattermost/action-mattermost-notify?tab=readme-ov-file#usage

saturninoabril (changes requested):
Thanks @mickmister, left comment.

.github/workflows/notify-mm-blog.yml
```diff
@@ -17,12 +17,11 @@ jobs:
         shell: bash
         run: |
           .github/workflows/notify-mm-blog.sh
-      - name: Notify channel 1
+      - name: Notify "Community Configuration" channel on community
         uses: mattermost/action-mattermost-notify@master
```
hanzei, line 21: Should we pin a version? That way we avoid pulling breaking changes.

hanzei (changes requested):

mickmister:
Closing @hanzei

PR #449, opened 2024-07-24
[fix] CI notify-community-server

phoinixgrr:
#### Summary

Aligning the action with the breaking changes introduced under https://github.com/mattermost/action-mattermost-notify/pull/13

Context: https://community.mattermost.com/core/pl/yb9ei7obg7ntjr483dbo1j8yoo

#### Ticket Link

 https://mattermost.atlassian.net/browse/CLD-8100

### DIFF
--- a/.github/workflows/notify-mm-blog.yml
+++ b/.github/workflows/notify-mm-blog.yml
@@ -18,11 +18,11 @@ jobs:
         run: |
           .github/workflows/notify-mm-blog.sh
       - name: Notify channel 1
-        uses: mattermost/action-mattermost-notify@master
-        env:
+        uses: mattermost/action-mattermost-notify@b7d118e440bf2749cd18a4a8c88e7092e696257a #v2.0.0
+        with:
           MATTERMOST_WEBHOOK_URL: ${{ secrets.MATTERMOST_WEBHOOK_URL_1 }}
       - name: Notify channel 2
-        uses: mattermost/action-mattermost-notify@master
-        env:
+        uses: mattermost/action-mattermost-notify@b7d118e440bf2749cd18a4a8c88e7092e696257a #v2.0.0
+        with:
           MATTERMOST_WEBHOOK_URL: ${{ secrets.MATTERMOST_WEBHOOK_URL_2 }}
 


toninis (approved):

.github/workflows/notify-mm-blog.yml
```diff
@@ -19,10 +19,10 @@ jobs:
           .github/workflows/notify-mm-blog.sh
       - name: Notify channel 1
         uses: mattermost/action-mattermost-notify@master
```
hanzei, line 21: Should we pin the version to avoid breaking changes?

phoinixgrr: Let's do that, rebasing and pushing the recommendation

phoinixgrr: @hanzei done https://github.com/mattermost/mattermost-marketplace/compare/3bfdb9c9353eb89a9790b02cc57476da839b44f8..84010f4ad0fd719a3ecb97af082fda213d980869

toninis (approved):

hanzei (approved):

phoinixgrr:
@hanzei Feel free to chery-pick this to all affected branches. `production`

hanzei:
Hm, this doesn't seem to have done the trick. https://github.com/mattermost/mattermost-marketplace/actions/runs/10353510589 reports an error: `You need to provide TEXT or PAYLOAD input`

@phoinixgrr Can you please take a look?

phoinixgrr:
> Hm, this doesn't seem to have done the trick. https://github.com/mattermost/mattermost-marketplace/actions/runs/10353510589 reports an error: `You need to provide TEXT or PAYLOAD input`
> 
> @phoinixgrr Can you please take a look?

@hanzei 
Ah yes. It needs a payload. What to actually send to the channels 
https://github.com/mattermost/action-mattermost-notify?tab=readme-ov-file#inputs

What would be that?

phoinixgrr:
Hmm I see. 

There is a bash script that needs to be alligned. https://github.com/mattermost/mattermost-marketplace/blob/master/.github/workflows/notify-mm-blog.sh

PR #452, opened 2024-08-13
[fix] CI Notification

phoinixgrr:
#### Summary

Inputing the output of `mattermost.json` to the PAYLOAD input of the notify action.

Builds on https://github.com/mattermost/mattermost-marketplace/pull/449#issuecomment-2286027219 

image


phoinixgrr:
> The changes look fine. Is there a way we can test them before merging?

Unfortunately not an easy one. Worst case scenario, notification will fail again. No other impact.

hanzei:
@mgdelacroix Would you be the right person from ABC to review this PR?

mgdelacroix:
@hanzei it seems that there is no specific knowledge around this in my team, so I'll do it

mgdelacroix (approved):

PR #454, opened 2024-08-28
Update enterprise label

esethna:
The MS Teams plugin is the only plugin in the marketplace to use this SKU label, so for accuracy this PR updates the help text and label to "Enterprise"


sbishel (approved):
LGTM

lieut-data (approved):

hanzei:
FWIW: Please note that Enterprise plugins will show up in the Marketplace even if a professional license is used. Changing the label might confuse customers using a professional licensee, as they will see the plugin and can install it, but it won't run, right?

PR #455, opened 2024-09-16
Release Calls v1.0.1

streamer45:
#### Summary

Releasing v1 on Marketplace. This is what's pre-packaged in MM v10.


lieut-data (approved):
This is probably fine, but I wonder if we should have merged `master` into `production` separately?

streamer45:
@lieut-data I admit we've been defaulting to `production` for a long time. I don't think Calls was ever ported to `master`.

cpoile (approved):

lieut-data:
> @lieut-data I admit we've been defaulting to production for a long time. I don't think Calls was ever ported to master.

Ah, maybe someone has updated `production` without regenerating `plugin.json` -- hence the noisy diffs. It's probably fine :p

PR #457, opened 2024-09-16
[MM-60556] Update Golang, server model, and lambda runtime

streamer45:
#### Summary

The first commit is just applying previously reviewed changes found in https://github.com/mattermost/mattermost-marketplace/pull/453

The second commit is migrating from the deprecated go1.x lambda runtime to a supported one (using `provided.al2023`). I have been following https://aws.amazon.com/blogs/compute/migrating-aws-lambda-functions-from-the-go1-x-runtime-to-the-custom-runtime-on-amazon-linux-2/ plus some Golang specific changes, most notably the need for the binary to be called `bootstrap` hence some packaging modifications.

This should also allow us to run the function on arm64 and likely reduce overall costs. I am not including this change in here, but it should be trivial to do later on.

So far I tested on a real lambda deployment using `make deploy-lambda`, which has been working okay. The idea is, once approved, to merge to `master`, make sure everything looks good in staging, and then port to `production`. After that, I'll try once more to add https://github.com/mattermost/mattermost-marketplace/pull/455 on top.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60556

PR #458, opened 2024-09-17
Update CI to work with new lambda artifact

streamer45:
#### Summary

The deploy step is currently [failing](https://github.com/mattermost/mattermost-marketplace/actions/runs/10904226898/job/30260222151) after https://github.com/mattermost/mattermost-marketplace/pull/457 went in because the expected file is not found. 

I don't really have a good way to test this, so I tried my best, but please let me know if you have any thoughts.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -11,7 +11,7 @@ jobs:
         shell: bash
     steps:
     - name: build/checkout-repo
-      uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       with:
         fetch-depth: 0
     - name: build/ensure-master-exists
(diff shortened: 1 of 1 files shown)


cpoile (approved):
No ideas, sorry. Hope it works!

.github/workflows/build-and-deploy.yml
```diff
@@ -50,14 +52,14 @@ jobs:
       AWS_DEFAULT_REGION: us-east-1
     steps:
     - name: build/checkout-repo
-      uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
toninis, line 55: This step is redundant .

toninis (approved):
It will work . May the force be with you 🚀

.github/workflows/build-and-deploy.yml
streamer45: let's remove it then

streamer45: @toninis On a second thought, I'll merge first, I need to send more PRs anyway otherwise as soon as I make a change here I'll invalidate reviews once more and start from scratch.

PR #466, opened 2024-11-12
Publish Calls v0.29.4 and v1.3.0

streamer45:
#### Summary

Publish latest Calls releases v0.29.4 and v1.3.0 to marketplace.


cpoile (approved):
That diff is a little wonky... Just want to double check that you're only adding the two? (it looks like it's editing older ones b/c of the bad diff)

streamer45:
@cpoile Yes, they always are a bit weird but I usually check the [whole file](https://github.com/mattermost/mattermost-marketplace/blob/57b0acd5b85a1f60945db562c16af05c597352ed/plugins.json) and all expected versions are there.

lieut-data (approved):

PR #474, opened 2025-01-21
Publish Calls v1.5.1 and v0.29.7

streamer45:
#### Summary

Publishing a couple of new dot releases for Calls


cpoile (commented):
What's happening in the diff? I'm seeing older releases getting their urls and checksums changed. Did github diff get confused?

streamer45:
@cpoile Same as https://github.com/mattermost/mattermost-marketplace/pull/466#pullrequestreview-2429741402 I suppose? I can check.

streamer45:
@cpoile All good. That said, the file could use some review because it probably doesn't make sense to include the entire history of releases but we may focus on just supported versions.

cpoile (approved):
👍

lieut-data (approved):

PR #478, opened 2025-02-19
Add v4.2.1 of mattermost-plugin-jira to the Marketplace

Kshitij-Katiyar:
#### Summary
This PR upgrades the [jira plugin](https://github.com/mattermost/mattermost-plugin-jira) to version [4.2.1](https://github.com/mattermost/mattermost-plugin-jira/releases/tag/v4.2.1).


plugins.json
```diff
@@ -28200,23 +28200,23 @@
...
-    "download_url": "https://plugins.releases.mattermost.com/release/mattermost-plugin-jira-v4.2.0.tar.gz",
-    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-jira/releases/tag/v4.2.0",
+    "download_url": "https://plugins.releases.mattermost.com/release/mattermost-plugin-jira-v4.2.1.tar.gz",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-jira/releases/tag/v4.2.1-rc1",
```
hanzei, line 28204: That seems wrong. Why are you replacing the existing version?

hanzei (changes requested):

plugins.json
Kshitij-Katiyar: @hanzei The last release for the Jira plugin i.e. `4.2.0` has been deleted from the Jira plugin since it had some issues so I replaced the existing version here.
Should we keep the existing version here?

wiggin77: @hanzei We must replace 4.2.0, which is a broken version.

wiggin77, line 28204: I don't think we should be putting an rc in marketplace.  Let's make a final release and update the PR here.

wiggin77 (changes requested):

plugins.json
Kshitij-Katiyar: @wiggin77 This is the final release 
the release notes URL is set to the RC link in all the releases
For example: 
```
 "repo_name": "mattermost-plugin-jira",
    "manifest": {
      "id": "jira",
      "name": "Jira",
      "description": "Atlassian Jira plugin for Mattermost.",
      "homepage_url": "https://github.com/mattermost/mattermost-plugin-jira",
      "support_url": "https://github.com/mattermost/mattermost-plugin-jira/issues",
      "release_notes_url": "https://github.com/mattermost/mattermost-plugin-jira/releases/tag/v4.1.1-rc1",
      "icon_path": "assets/icon.svg",
      "version": "4.1.1",
      "min_server_version": "7.8.0",
      "server": {
```

wiggin77 (dismissed):

plugins.json
hanzei: I think @wiggin77 is right. The "correct" tag exists, we should use it.
```suggestion
    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-jira/releases/tag/v4.2.1",
```

hanzei: Thanks for sharing that context :+1:

Kshitij-Katiyar: @hanzei @wiggin77 I have updated the release note URL

wiggin77 (approved):

hanzei (approved):

PR #482, opened 2025-03-20
Publish metrics plugin v0.6.0

streamer45:
#### Summary

Not sure why this wasn't published before today. Let me know if I am missing something.


cpoile (approved):

isacikgoz (approved):
> Not sure why this wasn't published before today. Let me know if I am missing something.

I think I just added this as pre-packaged so far. I don't remember if there was a specific reason to that.

sadohert:
@ewwollesen saw it published.  He has the screenshot.  But it 0.5.3.

streamer45:
> @ewwollesen saw it published. He has the screenshot. But it 0.5.3.

@sadohert Is it possible it was the pre-packaged version? It would still appear in the UX as if it was from Marketplace but in reality it isn't. There should be a LOCAL badge on the side.

image

PR #509, opened 2025-11-21
Add mattermost plugin jira v4.4.1

nevyangelova:
#### Summary
Add latest Jira version to the market place


.github/workflows/notify-mm-blog.yml
```diff
@@ -14,16 +14,19 @@ jobs:
         with:
           fetch-depth: 0
       - name: Generate webhook payload
+        id: notification
         shell: bash
         run: |
           .github/workflows/notify-mm-blog.sh
+          echo PAYLOAD=$(cat mattermost.json) >> ${GITHUB_OUTPUT}
```
marianunez, line 21: Are these changes intended? We should just be updating the Jira version

marianunez (changes requested):

marianunez:
We need to rebase to `production` branch as the diffs are too large and that is the branch we use to release

.github/workflows/notify-mm-blog.yml
ghost: @marianunez didn't change the base

marianunez: Ah ok, makes sense. It also fixed the large diff in the plugins file

marianunez (approved):

andrleite (approved):

PR #528, opened 2026-03-27
Add v1.0.0 of mattermost-plugin-wrangler to the Marketplace

nevyangelova:
#### Summary
Add latest wrangler plugin to the marketplace

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -36341,7 +36341,7 @@
       "id": "com.mattermost.wrangler",
       "name": "Wrangler",
       "description": "Manage messages across teams and channels",
-      "version": "",
+      "version": "1.0.0",
       "min_server_version": "10.11.0",
       "server": {
         "executables": {
--- a/plugins.json
+++ b/plugins.json
@@ -36327,10 +36327,10 @@
     "updated_at": "2025-01-10T13:57:06.235812976Z"
   },
   {
-    "homepage_url": "",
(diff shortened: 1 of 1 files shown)


plugins.json
```diff
@@ -36326,6 +36326,195 @@
     },
     "updated_at": "2025-01-10T13:57:06.235812976Z"
   },
+  {
+    "homepage_url": "",
+    "icon_data": "",
+    "download_url": "https://plugins.releases.mattermost.com/release/mattermost-plugin-wrangler-v1.0.0.tar.gz",
+    "release_notes_url": "",
```
avasconcelos114: We might want to set this to `https://github.com/mattermost/mattermost-plugin-wrangler/releases/tag/v1.0.0`

avasconcelos114: I think some of the missing info here may have been due to the wrangler plugin not being included in the `repositoryNames` on the `cmd/generator/main.go` file (also now wondering if the instances where we had to manually update some of these values were due to the same cause)

avasconcelos114 (commented):
Only have one thing to check prior to approval

ghost:
@avasconcelos114 thanks updated!

avasconcelos114 (approved):
LGTM! I'll create a tech-debt ticket for us to update the list of official plugins in the generation script

marianunez (approved):

PR #557, opened 2026-08-25
Updating missed URLs

avasconcelos114:
#### Summary

This PR is a follow up to the previous one as a couple of other URLs also needed to be fixed


avasconcelos114:
Turns out there are 2 `homepage_url` values in each marketplace object and these are the ones that actually get evaluated in the platform (no idea if the other ones are used at all)

avasconcelos114:
@coderabbitai review

nang2049 (approved):

